PR #146, opened 2015-03-13
Fixed some documentation around authenticateAccount

omgitstom:


robertjd:
Good to merge, failing tests in 0.8 is something we need to fix

PR #156, opened 2015-04-03
Updating deps

rdegges:
Updating all dependencies except for 'request' and 'nock', as other than those 2, all tests pass / things work as expected.


robertjd:
Hey!  So I was able to fix the issues with 0.8 via #157 

Could you create a new PR with the following changes:

* Branch off of master and PR into master.  We haven't been using develop, and it's behind.  The IT tests are failing on this branch because of it (run IT tests with `grunt it` and it uses API keys in `~/.stormpath/apiKey.properties` and needs to be able to create applications)
* Bring back 0.8 to travis

Thanks!

robertjd:
Also: if the 0.8 tests are failing with the new PR it means that the upgraded modules are not compatible with 0.8, in which case we would have to make an explicit decision to stop supporting 0.8 - and the PR should be titled accordingly so that we can collect attention.

NVM is good for trying out different versions of node while developing: https://github.com/creationix/nvm

PR #174, opened 2015-08-10
Oauth Password Grant Flow

robertjd:
This PR is implementing OAuth Authenticators with the new pattern, at the moment it only supports the password grant flow against our new /oauth API on applications.  I still need to implement the client credentials workflow.

I’d like to get some early feedback on the API and the return types (right now I just return the response we get from our new /oauth API).  See the tests for full examples, but here are the important ones RE the authentication API:

```javascript
/*
  OAuthAuthenticator - this is essentially the replacement
  for our previous "authenticateApiRequest" method, it expects
  to be given an HTTP request object.
*/

it('should be able to issue tokens for password grant requests', function(done){
  var authenticator = stormpath.OAuthAuthenticator(application);
  authenticator.authenticate({
    params:{
      grant_type: 'password'
    },
    body: {
      username: newAccount.username,
      password: newAccount.password
    }
  },function(err,result){
    common.assertPasswordGrantResponse(done)(err,result);
    passwordGrantResponse = result;
  });
});
```

```javascript
/*
  OAuthPasswordGrantRequestAuthenticator - if you want to use this grant type directly
*/

it('should create access tokens',function(done){
  var authenticator = new stormpath.OAuthPasswordGrantRequestAuthenticator(application);
  authenticator.authenticate({
    username: newAccount.username,
    password: newAccount.password
  },common.assertPasswordGrantResponse(done));
});
```


robertjd:
Ping @lhazlewood @omgitstom @rdegges

lhazlewood:
Why is there `params` and a `body` fields?  This is confusing if you understand OAuth - all three fields are part of the request 'body'.

Why not just have it like this?

```javascript
authenticator.authenticate({
    grant_type: 'password',
    username: newAccount.username,
    password: newAccount.password
}, errCallback);
```

Based on the `grant_type`, you will always know what other fields to look for.  I'm curious as to why they need to be wrapped as most other OAuth implementations don't do that.  Thoughts?

robertjd:
`OAuthAuthenticator` expects an Http Request, what you're seeing is the common data structure for http requests in Node

lhazlewood:
Ok, that makes more sense, but now I have to ask a follow up question:

`grant_type` is not a request parameter.  All three of those fields in an OAuth request are `application/x-www-form-urlencoded` body parameters - why are they represented as 2 different things in ours?

lhazlewood:
Or maybe a better question is:

The OAuth Authenticator takes in a request.   So why aren't we just relaying the Node request directly to our function?

lhazlewood:
(that was the intention of accepting a request object - so the caller can directly relay the request object without the caller having to re-create any intermediate object/state).

robertjd:
Ooof.. I totally merged this by accident, I'm reverting.  Anyways..

Yes, the grant type should be in the body - no idea why I did that :)

And yes, you'll be able to pass the request object directly, assuming you have an prior filter which has parsed the body data for you (anyone using node is probably using the [body-parser](https://github.com/expressjs/body-parser) or something similar, so this is just a matter of noting it in the documentation)

PR #182, opened 2015-08-26
Cleaning up require statements to be consistent.

rdegges:
I'm organizing them so that:

- We always have 'use strict';
- We have node imports first.
- We have third party imports second.
- We have local imports third.
- All are in alphabetical order.

Tests all pass.


lhazlewood:
Question: if you're making policy decisions about how code is organized (which is fine - this makes sense to me) how do you/we ensure that this is maintained moving forward?  I use IntelliJ, and I can tell it to automatically arrange imports as defined above, but how do you ensure the entire team does this?

I ask because these kinds of changes mostly add version control 'noise' and don't actually impact functionality, so if there's no guarantee that this stays consistent (e.g. w/ an IDE), this *will* diverge in the future, and there will be more noise.  It then becomes busywork maintenance which is a time sink and not worth it...

rdegges:
Welp, this is how people do it in python / other languages by hand. It doesn't seem uncommon to me?

This commit brings everything into order, so should be easier to maintain over time. What I like to do when coding is fix small style issues as I'm going: always try to leave the codebase a little better off than it was when you touched it ^^

PR #183, opened 2015-08-26
Convenience Configuration

rdegges:
This commit modifies the configuration client to do a little bit of magic.

The idea is this: right now, it is quite tricky to enable the default stuff in Express for websites: login / logout / register.

You have to know about the configuration options, and specify an ugly nested config that is quite lengthy and eats a lot of space in your code.

What this PR does is allows users to simplify their configuration through 2 optional options: `website` and `api`.

If you say:

```
var client = new Client({
  website: true
});
```

Then you'll get the registration / login / logout functionality all enabled with our default settings (like the old express-stormpath behavior).

If you say:

```
var client = new Client({
  api: true
});
```

You'll get our oauth endpoint enabled by default, eg: `/oauth/token` -- this is ideal for users building any sort of API service.

@robertjd and I have discussed this previously and think it's a good idea.

### DIFF
--- a/lib/Config.js
+++ b/lib/Config.js
@@ -59,6 +59,26 @@ function Config(options){
     this.client.apiKey.secret = props.secret;
   }
 
+  // If a user enables a boolean configuration option named `website`, this
+  // means the user is building a website and we should automatically enable
+  // certain features in the library meant for users developing websites.  This
+  // is a simpler way of handling configuration than forcing users to specify
+  // all nested JSON configuration options themselves.
+  if (this.website) {
+    this.web.register.enabled = true;
+    this.web.login.enabled = true;
+    this.web.logout.enabled = true;
+  }
+
+  // If a user enables a boolean configuration option named `api`, this means
+  // the user is building an API service, and we should automatically enable
+  // certain features in the library meant for users developing API services --
+  // namely, our OAuth2 token endpoint (/oauth/token).  This allows users
+  // building APIs to easily provision OAuth2 tokens without specifying any
+  // nested JSON configuration options.
+  if (this.api) {
+    this.web.oauth2.enabled = true;
+  }
 }
 
 Config.prototype.getEnvVars = function(){
--- a/test/sp.client_test.js
+++ b/test/sp.client_test.js
@@ -84,6 +84,39 @@ describe('Client', function () {
       done();
     });
 
+    it('should enable all web routes if the website config option is specified', function(done) {
+      var client = new Client({
+        client: {
+          apiKey: {
+            id: '1',
+            secret: '1'
+          }
+        },
+        website: true
+      });
+
+      assert(client.config.web.register.enabled);
+      assert(client.config.web.login.enabled);
+      assert(client.config.web.logout.enabled);
+
+      done();
+    });
+
+    it('should enable all oauth2 routes if the api config option is specified', function(done) {
+      var client = new Client({
+        client: {
+          apiKey: {
+            id: '1',
+            secret: '1'
+          }
+        },
+        api: true
+      });
+
+      assert(client.config.web.oauth2.enabled);
+      done();
+    });
+
     it('should emit a ready event with the client itself as the value', function(done) {
       var client = new Client({
         client: {


lhazlewood:
We need to be careful here - all config should be the same across all integrations as much as is possible.  If it is config related, it needs to be captured in the SDK Spec work in the other GitHub repo.

robertjd:
Agreed, we'll get this copied over there as well

lhazlewood:
I'm actually thinking this isn't necessary - all of our endpoints should be enabled by default, and they can turn off the ones they don't want.

Why add this option?  Also, what if you're both a website and an api?

Also, the Client shouldn't know *anything* about URLs/routes, etc.  A Client is a client to the REST API, and shouldn't have any knowledge of things like framework integrations.

In other words, the framework integrations 'sit on top of' or 'delegate' to the client.  The Client shouldn't know anything about URL options or controllers, etc.  I'm confused why this is in the Client configuration...

rdegges:
So, few things:

After talking with a lot of users, I think it's a bad idea to have this stuff on by default: it's very very confusing / concerning to knowledgable users. They don't understand why web pages will be enabled by default that allow users to be created / etc. -- it's sorta scary to people building APIs or services that use Stormpath data, but aren't necessarily front-end services.

I've gotten this a bunch, which makes me think turning both off by default is a good idea.

With that said, enabling things is harder than necessary right now. So I vote for including these convenience methods which are intuitive to users, and easily enable our 'defaults'. If you're building both a website AND api in the same project, you could just enable both flags and be fine.

Regarding why this logic is in the Node SDK and not the Express SDK: all the framework integrations will have a common dependency on the Node SDK, so it makes sense so have the configuration builder there. All it does is parse / pull config options, it doesn't actually *do* anything.

The Express library is still 100% responsible for doing stuff with these config options. We just thought that since the underlying library is receiving the options, it makes sense to make it responsible for parsing / validating all configuration options in a single place.

lhazlewood:
I feel like I'm missing something here, or this is some weird node-specific thing...

It usually doesn't make sense to do it all in one place - it makes sense to do it in the place where the options are relevant, otherwise you conflate the responsibility of the Client to do things beyond its design (this is the Single Responsibility Principle).  We've seen this problem with authentication off of the application (various authentication mechanisms bloated the application's dependencies and logic).

I think we need to have a call so I can understand what is being proposed.

robertjd:
We need a common, re-usable place to put the config parser.  The alternative is to create a new node module (which feels like overkill).  The client constructor is a common entry point for all things stormpath + node, so it really makes the most sense here.  If we change our minds, this is a behind-the-api implementation detail that can change later without affecting how developers use our node software (which always starts with a `new stormpath.Client(Config)` invocation)

lhazlewood:
I think I'm starting to understand.  How does the express stuff access the config and turn things on or off?

robertjd:
The constructed config object is accessible via `new Stormpath.client(config}).config`.  The express library expects this object to exist, and then uses it to decide which features to wire up (and how).

For example:

https://github.com/stormpath/stormpath-express/blob/angular-integration/lib/stormpath.js#L148-L158

lhazlewood:
That the client knows about the config (i.e. has a config property) is an inverted way of looking at things IMO - the client should 'know' only about the things it has dependencies on and nothing extra.

In other words, Config sits 'above' the Client - the client doesn't need to know about config at all.  Config is more 'meta' than the client.  For example, instead of tightly coupling the Client to config, this makes more sense (and is better practice with the SRP principle):

1.  Read in a config.
2.  Based on the config, create the things that the config reflects (a data store, a client, a cache manager, etc).
3.  Wire these things together (client uses a cache manager, etc)
4.  Return a client instance based on the dependencies being wired.

All of these things are completely orthogonal to the Client - a client shouldn't know about any of this stuff.  More concretely creational patterns and usage patterns should (generally) always be separate because lifecycle behavior is totally unrelated to post-creation behavior.

The current design requires all of the above logic to be tightly coupled to the client implementation, which makes for brittle code and unnecessary Client implementation changes if config changes.

Does this make sense?

rdegges:
Hey Les,

You're totally right! We actually did this a bit here: https://github.com/stormpath/stormpath-sdk-node/blob/master/lib/Config.js

We have a Config object, we just map it to the Client for convenience sake (when you create a Client, you can access config via client.config).

Since we can refactor this after release, I'm going to merge this in right now so we can move forward. There's still a lot of cleanup to do in this library in regards to architecture / best practices -- there are quite a few issues that myself and Robert find annoying atm.

We can discuss this more on Friday in person, and create issues here for things we still need to update.

Thanks <3

PR #205, opened 2015-09-21
fix getAccessToken expires too quickly

### LINKED ISSUE #204: getAccessToken() default TTL value is incorrectly set as 3600ms (expected 3600s)
When `getAccessToken()` method is called, the `expires_in` value is incorrectly set.
This is due to `nJwt.setExpiration()` expects values in _milliseconds_. Not _seconds_

Also noticed that `jwt.ttl` value is not set by nJwt library.

ck-lee:
closes #204 
A quick fix by setting `setExpiration()` value correctly


robertjd:
Closing in favor of #214 - thank you for reporting this bug!

PR #212, opened 2015-09-30
Implement getters for access tokens and refresh tokens

robertjd:
Implemented these functions:

account.getAccessTokens()
account.getRefreshTokens()
client.getAccessToken()

The modifications to the account_it.js test ended up causing
some problems with the application and organization tests, so 
I have patched those to be more reliable

Created a fixture so that we can share test setup between the account
test and client test (both need to create an access token for an 
account)


test/fixtures/account-token.js
```diff
@@ -0,0 +1,69 @@
...
+}
+AccountAccessTokenFixture.prototype.before = function before(done) {
+  var self = this;
+  helpers.getClient(function(_client){
+
+    self.client = _client;
+    helpers.createApplication(function(err,app){
+      if (err) {
+        return done(err);
+      }else{
```
timothyej: You can save some indentation by removing this `else`, as there is not further code after this statement, and you're doing a `return` just before it.

test/fixtures/account-token.js
```diff
@@ -0,0 +1,69 @@
...
+        self.application = app;
+        helpers.getDefaultAccountStore(app,function(err,dir){
+          if (err) {
+            return done(err);
+          }else{
```
timothyej: You can also save some indentation here by removing this `else` :)

test/fixtures/account-token.js
```diff
@@ -0,0 +1,69 @@
...
+            });
+          }
+        });
+
+      }
+    });
+  });
+};
+AccountAccessTokenFixture.prototype.after = function after(done) {
+  var self = this;
```
timothyej: A see a lot of these `self`. Why not create a variable for what you need instead? :) Like this:

```javascript
var directory = this.directory;

this.application.delete(function(err){
  if (err) {
    return done(err);
  }
  directory.delete(done);
});
```

What do you think?

lib/resource/Account.js
```diff
@@ -11,6 +11,24 @@ function Account(data) {
 }
 utils.inherits(Account, DirectoryChildResource);
 
+Account.prototype.getAccessTokens = function getAccessTokens(/* [options,] callback */) {
+  var self = this;
+  var args = Array.prototype.slice.call(arguments);
+  var callback = args.pop();
+  var options = (args.length > 0) ? args.shift() : null;
+
+  return self.dataStore.getResource(self.accessTokens.href, options, require('./AccessToken'), callback);
```
timothyej: You don't need the `self` here? :)

test/fixtures/account-token.js
robertjd: Personal style, I'm okay with switching to no-else block.  I think @rdegges prefers that style?

test/fixtures/account-token.js
robertjd: I do like this style, it works for the `after` method.  Do you know how to make it work with the `before` method, since we assign an object to the null value at a later time?

lib/resource/Account.js
robertjd: No, this is copy and paste mess.  I want to change this everywhere we can.

PR #213, opened 2015-10-03
Refactor: Configuration

typerandom:
Changes the way that configuration is loaded by using the (new) [stormpath-config](https://github.com/stormpath/stormpath-node-config) module and its strategies.

This new way of loading configuration breaks the loading phase down into several configurable steps. Not only is this easier to test and manage. This also makes the configuration process extendable from other modules (see [express-stormpath branch refactor-client-config](https://github.com/stormpath/stormpath-express/tree/refactor-client-config)).

#### What this actually fixes

* **YAML supported.** There can now be both `~/.stormpath/stormpath.[json/yml]` and `./stormpath.[json/yml]` files. Also the default configuration (previously config.json) is now all YAML.
* **Fully async.** Previously the configuration was both sync (blocking fs calls) and async. With most of the configuration loaded sync during Client instantiation and some enrichment running after async and eventually triggering the 'ready' event. This made the client a bit inconsistent since one could both receive an exception and then errors through the 'error' event. Also, the client could be instantiated and used without having emitted its 'ready' event. Now, the 'ready' event will first be emitted once the configuration is fully loaded. Though, in order for users to still be able to just instantiate the client and call the API without waiting for the 'ready' event the client will be wrapped in a proxy and keep all calls to the Client waiting until the client is actually ready.

* **Only client config.** Previously the SDK contained configuration related to the express integration. This has now been moved into the [express-stormpath](https://github.com/stormpath/stormpath-express/blob/refactor-client-config/lib/client.js) module.

* **Testable.** Previously the configuration was a bit shattered and hard to test. Now it's separated into isolated strategies so testing can be done both separately (single strategy) and many together (with ConfigLoader, grouping of strategies).

(additional)

* **'Ready' event only fired once.** Previously the ready event was only fired once. So if someone attached to the client too late (after the event had been emitted), then they would never receive it. This is now fixed, so that if one attach to the 'ready' event, and the client is ready, then the callback is called immediately.

#### How to verify

1. Check out the `refactor-config-loader` branch.
2. Run `$ npm test` and hope for the best!
3. Run `npm link` to link this branch globally so that it's used when the `stormpath` module is referenced.
4. Create a new empty project (new directory).
5. Run `npm link stormpath` so that this project links any `stormpath` includes to our `refactor-config-loader` branch.
4. Remove any Stormpath environment variables.
3. Create a new file (e.g. `app.js`) and initialize a new stormath Client without providing any configuration (i.e. `new stormpath.Client()`).
4. Add a call to the `client.getApplications` method.
5. Run the application and verify that you get an error stating `Error: API key ID and secret is required`.
6. Create a `stormpath.(json or yml)` file in your `.stormpath` home folder. Add your Stormpath credentials.
7. Run the application and verify that you're able to retrieve the account.
8. Create a `stormpath.(json or yml)` file in the same directory as your application. Add some configuration that overwrites the configuration you specified in your home directory (step 3).
9. Run the application and verify that the new configuration you set has affected your client.
10. Add configuration to your client code (i.e. `new stormpath.Client({/* config */})`).
11. Run the application and verify that the configuration has overridden the previous configurations set (in step 3 and 8).

(optional/additional)

If you already have an existing application, then upgrade to the new `stormpath` module, then test the application. It should function exactly the same as before (except from the `express-stormpath` module, since this will have to be updated to work with the new configuration loader. See [stormpath-express/refactor-client-config](https://github.com/stormpath/stormpath-express/tree/refactor-client-config)).

(optional/additional)

Create a new application according to [the SDK tutorial](http://docs.stormpath.com/nodejs/quickstart/) and test that the application works as previously expected.


rdegges:
Very cool! This looks so great, thanks so much!

I only have one minor change I'd like to see, which is to get rid of the whitspace before function params per our new style stuff, eg: `function ()` -> `function()`.

+1

lib/Client.js
```diff
@@ -55,207 +76,36 @@ util.inherits(Client, Object);
...
-            return cb(null,null);
-          }
-        } else {
-          return cb(null,null);
-        }
-      });
-    });
-  }
+Client.prototype.on = function() {
+  var args = Array.prototype.slice.call(arguments);
```
timothyej: Any reason for why you're doing this instead of just using `arguments` directly?

typerandom:
@rdegges Thanks!

And that's really interesting since it's one of the things that both me and Timothy have been discussing. I think there should always be a space after a keyword. I.e:

`function () { ... }` and `function Foo() {}` is ok, but not `function Foo () { ... }` or `function(){ ... }`.

This is also according to Crockfords JS conventions:

>If a function literal is anonymous, there should be one space between the word function and the ( (left parenthesis). If the space is omitted, then it can appear that the function's name is function, which is an incorrect reading.

http://javascript.crockford.com/code.html

What do you think about that?

lib/Client.js
typerandom: Because I think it's clearer to work against an array (and a copy) than it is to work with the [arguments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) object. It's also easier to make errors when working against the arguments object, because most assume it's just an array, when it's not. Also, *args* is shorter and looks better to me, than having to retype *arguments*.

lib/proxy/ObjectCallProxy.js
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Attach onto the source object and intercept all calling
+ * methods with callbacks (where last argument is function).
+ *
+ * @param fn [predicateFn] - Predicate to filter methods by (optional).
+ */
+ObjectCallProxy.prototype.attach = function (predicateFn) {
+  var outerScope = this;
+  var source = outerScope.source;
```
timothyej: `outerScope` is pretty generic and in this case actually *this* scope, and not the outer. Instead, I would declare two variables here, `attached` and `pending`:

```javascript
var attached = this.attached;
var pending = this.pending;
var source = this.source;
...
```

Then I replace the use of `outerScope.pending` to only use `pending` etc.

What do you think?

lib/Client.js
robertjd: I also prefer this `args` approach, for the reasons mentioned.  It makes it much easier to work with complex argument combinations.

lib/proxy/ObjectCallProxy.js
typerandom: @timothyej I agree, this makes no sense here, so changing it :)

lib/proxy/ObjectCallProxy.js
```diff
@@ -0,0 +1,116 @@
...
+ * @param {Error} [err] - Error to release all callbacks with (optional).
+ */
+ObjectCallProxy.prototype._release = function (err) {
+  var source = this.source;
+  var pending = this.pending;
+
+  this.pending = [];
+
+  pending.forEach(function (call) {
+    var args = call.args,
```
timothyej: Any reason for why you are using joined declarations here and not anywhere else? ;)

typerandom: I was just trying out that style since we were discussing it at that time, but forgot to remove it :) Fixing it.

.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+insert_final_newline = true
+indent_size = 2
+
+[**.yml]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
timothyej: How ironic, this file doesn't include a `final_newline` itself ;)

typerandom: Not really. It's an `.editorconfig` file :)

lib/Client.js
```diff
@@ -24,28 +23,50 @@ function Client(config) {
...
   // make unnecessary API requests if this object is looked up more than once.
   self._currentTenant = null;
 
-  self.mergeRemoteConfig(function(err) {
-    return err ? self.emit('error', err): self.emit('ready', self);
+  // Indicates whether or not this client is ready yet.
+  self._isReady = false;
+
+  // Setup how we load our configuration.
+  var configLoader = null;
```
timothyej: Why are you setting this to `null`? Variables will be `undefined` if not assigned. Besides, you're still assigning it in the block below.

lib/Client.js
timothyej: :+1:

lib/Client.js
typerandom: @timothyej As a convention. I always do that, since I know that I'll be able to type check it against *NULL* if it isn't set.

robertjd: +1

PR #228, opened 2015-10-13
Feature: Load config env after config files.

### LINKED ISSUE #225: Environment config should be loaded after files
As discussed [here](https://github.com/stormpath/express-stormpath/issues/135) we should load the environment config after we've loaded config from files. I.e. so that the config load order is:

1. File `~/.stormpath/apiKeys.properties`.
2. File `~/.stormpath/stormpath.json`.
3. File `(application path)/apiKeys.properties`.
4. File `(application path)/stormpath.json`.
5. Environment.
6. Config provided to integration (i.e. `stormpath.init(app, { /* here */ });`).

typerandom:
Changes so that environment config is loaded after config files.

##### How to verify

Use an existing application that integrates with the `stormpath` module, and test that the Stormpath environment config (`STORMPATH_`) overrides Stormpath config set in files (`~/.stormpath/stormpath.json`, `~/.stormpath/apiKey.properties`, `./stormpath.json`, `~/.stormpath/apiKey.properties`). And that configuration passed directly to the client overrides that in the environment.

I.e. the config load order should now be:

1. File `~/.stormpath/apiKeys.properties`.
2. File `~/.stormpath/stormpath.json`.
3. File `(application path)/apiKeys.properties`.
4. File `(application path)/stormpath.json`.
5. Environment.
6. Config provided to integration (i.e. `stormpath.init(app, { /* here */ });`).

Fixes #225.

### DIFF
--- a/test/sp.config.configLoader_test.js
+++ b/test/sp.config.configLoader_test.js
@@ -8,7 +8,6 @@ var FakeFS = require('fake-fs');
 var common = require('./common');
 var assert = common.assert;
 
-var stormpathConfig = require('stormpath-config');
 var configLoader = require('../lib/configLoader');
 
 describe('Configuration loader', function () {
@@ -70,8 +69,9 @@ describe('Configuration loader', function () {
     afterIt.push(restoreEnv);
 
     loader.load(function (err, config) {
-      assert.isNotNull(err);
+      assert.isUndefined(config);
 
+      assert.isNotNull(err);
       assert.equal(err.message, 'API key ID and secret is required.');
 
       done();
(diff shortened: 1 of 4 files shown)


test/sp.config.configLoader_test.js
```diff
@@ -52,9 +51,16 @@ describe('Configuration loader', function () {
     }
   });
 
-  it('should load without configuration', function(done) {
+  it('should load configuration with only api key', function(done) {
```
robertjd: It's unclear to me what this test is asserting (maybe it needs a more descriptive statement).  Other than that I think this PR looks great.

typerandom: I fully agree. The thing is that this test did not fully work before, since it actually got configuration from env. But I'm instead changing the test. Will have an update to this in a few minutes.

typerandom:
@robertjd Awesome!

Who wants to push the `merge` button? ;)

robertjd:
Pushed :)

PR #240, opened 2015-10-16
Fix: Pass local validation from OAuthAuthenticator to JwtAuthenticator

### LINKED ISSUE #216: OAuthAuthenticator does not propagate info for local validation to JwtAuthenticator
On lib/oauth/authenticator.js, the information for local validation is not correctly propagated to JwtAuthenticator.

typerandom:
Fixes so that the local validation option is passed from OAuthAuthenticator to JwtAuthenticator.

Closes #216.

### DIFF
--- a/lib/jwt/jwt-authenticator.js
+++ b/lib/jwt/jwt-authenticator.js
@@ -45,7 +45,8 @@ JwtAuthenticator.prototype.authenticate = function authenticate(token,cb){
       if(self.localValidation){
         cb(null, new JwtAuthenticationResult(self.application,{
           jwt: token,
-          expandedJwt: jwt
+          expandedJwt: jwt,
+          localValidation: true
         }));
       }else if(jwt.header.kid){
         // If the KID exists, this was issued by our API
--- a/lib/oauth/authenticator.js
+++ b/lib/oauth/authenticator.js
@@ -29,13 +29,21 @@ OAuthAuthenticator.prototype.authenticate = function authenticate(req,callback){
   }else{
     throw new Error('authenticate must be called with a request object and callback function');
   }
+
   var authHeader = req && req.headers && req.headers.authorization;
 
   if (authHeader){
     if(authHeader.match(/Bearer/i)){
+      var authenticator = new JwtAuthenticator(self.application);
+
+      if (this.localValidation) {
+        authenticator.withLocalValidation();
+      }
+
       var parts = authHeader.split(' ');
       var token = parts[parts.length-1];
-      new JwtAuthenticator(self.application).authenticate(token,done);
+
+      authenticator.authenticate(token, done);
     }else{
       done(self.unauthenticated());
     }
--- a/lib/oauth/authenticator.js
+++ b/lib/oauth/authenticator.js
@@ -5,6 +5,16 @@ var JwtAuthenticator = require('../jwt/jwt-authenticator');
 var OAuthPasswordGrantRequestAuthenticator = require('../oauth/password-grant').authenticator;
 var OAuthRefreshTokenGrantRequestAuthenticator = require('../oauth/refresh-grant').authenticator;
 
+function retrieveAuthTokenFromRequest(req) {
+  var authHeader = req && req.headers && req.headers.authorization;
+
+  if (!authHeader || !authHeader.match(/Bearer/i)) {
+    return false;
+  }
+
+  return authHeader.split(' ').pop();
+}
+
 function OAuthAuthenticator(application) {
   if (!(this instanceof OAuthAuthenticator)) {
     return new OAuthAuthenticator(application);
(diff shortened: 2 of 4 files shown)


lib/oauth/authenticator.js
```diff
@@ -29,13 +29,21 @@ OAuthAuthenticator.prototype.authenticate = function authenticate(req,callback){
...
 
   if (authHeader){
     if(authHeader.match(/Bearer/i)){
+      var authenticator = new JwtAuthenticator(self.application);
+
+      if (this.localValidation) {
+        authenticator.withLocalValidation();
+      }
+
       var parts = authHeader.split(' ');
```
timothyej: Would be nice to have these declarations below `var authenticator ...` instead (on line 38) :)

typerandom: That I disagree with. I believe more in having locality than "all declarations at top". Unless you have a really good reasoning for it:)

timothyej: I didn't mean at the top of this scope, just 4 lines up :)

I find it easier to read because I can see that we have 3 variables; `authenticator`, `parts`, and `token` for this code block, and they are all used together, right? *Setup* and then *execution*...

```javascript
var authenticator = new JwtAuthenticator(self.application);
var parts = authHeader.split(' ');
var token = parts[parts.length-1];

if (this.localValidation) {
  authenticator.withLocalValidation();
}

authenticator.authenticate(token, done);
```

**vs**

```javascript
var authenticator = new JwtAuthenticator(self.application);

if (this.localValidation) {
  authenticator.withLocalValidation();
}

var parts = authHeader.split(' ');
var token = parts[parts.length-1];

authenticator.authenticate(token, done);
```

typerandom: I actually think that makes it harder to read, since you have to grasp the whole block of code at once. And it's actually two things. First it enables validation (if that is set) then it extracts tokens from a header and tries to authenticate.

typerandom: @timothyej Refactored it now for readability. See 9fa3d696ed3b86769e0ef4636328d1aca53ba8d0.

timothyej: Yaaaay! :+1:

lib/oauth/authenticator.js
```diff
@@ -21,31 +31,37 @@ OAuthAuthenticator.prototype.withLocalValidation = function withLocalValidation(
...
+      authenticator.withLocalValidation();
     }
-  }else if (req.body && req.body.grant_type==='password'){
-    new OAuthPasswordGrantRequestAuthenticator(self.application).authenticate(req.body,callback);
-  }else if (req.body && req.body.grant_type==='refresh_token'){
-    new OAuthRefreshTokenGrantRequestAuthenticator(self.application).authenticate(req.body,callback);
-  }else{
-    done(self.unauthenticated());
+  } else if (req.body) {
+    token = req.body;
```
timothyej: Add a new line break after this line.

typerandom: There is one? :) ;);)

lib/oauth/authenticator.js
```diff
@@ -21,31 +31,37 @@ OAuthAuthenticator.prototype.withLocalValidation = function withLocalValidation(
...
+    if (req.body.grant_type === 'password') {
```
timothyej: This is a perfect use-case for a `switch` statement :)

```javascript
switch (req.body.grant_type) {
  case 'password':
    authenticator = new OAuthPasswordGrantRequestAuthenticator(self.application);
    break;

  case 'refresh_token':
    authenticator = new OAuthRefreshTokenGrantRequestAuthenticator(self.application);
    break;
}
```

typerandom: Yeah, I actually added that and then removed it. Because felt like it bloated more than it added value.

timothyej: I think it's easier to follow the logic of these lines rather than with the long `if-else` lines:

```javascript
switch (req.body.grant_type) {
  case 'password':
    authenticator = new OAuthPasswordGrantRequestAuthenticator(self.application);
    break;

  case 'refresh_token':
    authenticator = new OAuthRefreshTokenGrantRequestAuthenticator(self.application);
    break;
}
```

**vs**

```javascript
if (req.body.grant_type === 'password') {
  authenticator = new OAuthPasswordGrantRequestAuthenticator(self.application);
} else if (req.body.grant_type === 'refresh_token') {
  authenticator = new OAuthRefreshTokenGrantRequestAuthenticator(self.application);
}
```

timothyej: But it's your decision :+1:

typerandom: Yeah. I personally think it bloats more than it actually gives value. I'm all for using `switch` but when it actually removes lines and adds value. In this case I'd say that it doesn't since it's uncertain exactly how things will be resolved. I.e. in the future there might be some other `req.body`-value and then the `switch` clause will be very confusing since it might imply that all it should handle is `grant_type`´s. Whereas now it's more open to interpretation.

timothyej: :+1:

typerandom:
@timothyej I'm a little bit confused. Do you think you could look at this? https://coveralls.io/builds/3917635/source?filename=lib%2Foauth%2Fauthenticator.js

It appears in the coverage report that almost none of the lines are being covered for. But they are plenty of tests that should cover most of it:

https://github.com/stormpath/stormpath-sdk-node/blob/feature-oauth-authenticator-does-not-pass-local-validation/test/it/oauth_authenticator.js

Am I looking on something the wrong way or is there something strang going on?

timothyej:
@typerandom What was the test coverage at before this PR and with this PR? Is their number `91.44%` correct or incorrect?

typerandom:
It looks like the correct number. But the thing I'm having trouble with is the coverage reporting. I'd like to see which lines aren't being covered for, but the report shows almost all as red... And that to me is really confusing.

typerandom:
Finally found the issue! The oauth authenticator tests weren't run because it wasn't discovered! The simple fix was to rename `oauth_authenticator.js` to `oauth_authenticator_it.js`. :)

lib/jwt/jwt-authenticator.js
```diff
@@ -45,7 +45,8 @@ JwtAuthenticator.prototype.authenticate = function authenticate(token,cb){
       if(self.localValidation){
         cb(null, new JwtAuthenticationResult(self.application,{
           jwt: token,
-          expandedJwt: jwt
+          expandedJwt: jwt,
+          localValidation: true
```
robertjd: Nice touch, I can see how this would be useful.

robertjd:
Looks good, merging!

PR #267, opened 2015-10-30
Feature: Speed-up tests by caching client

typerandom:
Execute tests faster by only instantiating one client when using the `helper.getClient()` helper method.

### DIFF
--- a/test/it/client_it.js
+++ b/test/it/client_it.js
@@ -16,19 +16,16 @@ describe('Client', function() {
   describe('creation', function() {
     it('should not throw', function(done) {
       assert.doesNotThrow(function () {
-        helpers.loadApiKey(function(apiKey) {
-          var client = new stormpath.Client({
-            apiKey: apiKey,
-            skipRemoteConfig: true
-          });
+        var client = new stormpath.Client({
+          skipRemoteConfig: true
+        });
 
-          client.on('error', function (err) {
-            throw err;
-          });
+        client.on('error', function (err) {
+          throw err;
+        });
 
-          client.on('ready', function () {
-            done();
-          });
+        client.on('ready', function () {
+          done();
         });
       });
     });
--- a/test/it/helpers.js
+++ b/test/it/helpers.js
@@ -7,44 +7,26 @@ var common = require('../common');
 var uuid = common.uuid;
 var stormpath = common.Stormpath;
 
+var loadedClient = null;
 var hasCleanupRun = false;
 var testRunId = uuid.v4().split('-')[0];
 
-function loadApiKey(cb) {
-  var id = process.env.STORMPATH_CLIENT_APIKEY_ID;
-  var secret = process.env.STORMPATH_CLIENT_APIKEY_SECRET;
-  var homeDir = process.env[(process.platform === 'win32' ? 'USERPROFILE' : 'HOME')] || '';
-  var apiKeyFilePath = homeDir + '/.stormpath/apiKey.properties';
-
-  if (id && secret) {
-    return process.nextTick(function() {
-      cb(new stormpath.ApiKey(id, secret));
-    });
+function getClient(cb) {
+  if (loadedClient) {
+    return cb(loadedClient);
   }
 
-  stormpath.loadApiKey(apiKeyFilePath, function(err, apiKey) {
-    if (err) {
-      throw err;
-    }
-
-    cb(apiKey);
+  var client = new stormpath.Client({
+    skipRemoteConfig: true
   });
-}
 
-function getClient(cb) {
-  loadApiKey(function(apiKey) {
-    var client = new stormpath.Client({
-      apiKey: apiKey,
-      skipRemoteConfig: true
-    });
+  client.on('error', function (err) {
+    throw err;
+  });
 
-    client.on('error', function (err) {
-      throw err;
-    });
-
-    client.on('ready', function () {
-      cb(client);
-    });
+  client.on('ready', function () {
+    loadedClient = client;
+    cb(client);
   });
 }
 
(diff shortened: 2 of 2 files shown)


test/it/helpers.js
```diff
@@ -2,31 +2,13 @@ var common = require('../common');
...
-
-    cb(apiKey);
-  });
-}
+var loadedClient = null;
 
 function getClient(cb) {
-  loadApiKey(function(apiKey) {
+  if (loadedClient) {
+    cb(loadedClient);
```
timothyej: You could return here and also save some indentation by not having the `else` statement below.

typerandom: Yes, though I feel that from a readability perspective, it's much clearer to follow.

test/it/client_it.js
```diff
@@ -16,19 +16,16 @@ describe('Client', function() {
...
-          });
+        var client = new stormpath.Client({
+          skipRemoteConfig: true
+        });
+
+        client.on('error', function (err) {
+          throw err;
+        });
+
+        client.on('ready', function () {
```
timothyej: If the `ready` event sends an error as the first argument, it should be passed on to the `done` callback:

```javascript
client.on('ready', done);
```

Or:

```javascript
client.on('ready', function (err) {
    done(err);
});
```

typerandom: Why? In this case throwing an exception makes more sense. Consider this, the client readies, then when calling something, a client error happens. And by doing what you suggest it's not certain it will be handled, since the `done()` callback is just expected to be called once. Whereas any exception being thrown will be picked up properly by mocha.

test/it/helpers.js
timothyej: I think it's bad for readability to encapsulate almost an entire function inside an `else` statement:

```javascript
function getClient(cb) {
  if (loadedClient) {
    cb(loadedClient);
  } else {
      skipRemoteConfig: true
    });

    client.on('error', function (err) {
      throw err;
    });

    client.on('ready', function () {
      loadedClient = client;
      cb(client);
    });
  }
}
```

**vs.**

```javascript
function getClient(cb) {
  if (loadedClient) {
    return cb(loadedClient);
  }

    skipRemoteConfig: true
  });

  client.on('error', function (err) {
    throw err;
  });

  client.on('ready', function () {
    loadedClient = client;
    cb(client);
  });
}
```

timothyej: Your scenario:

* If the client is loaded, invoke the callback with the loaded client
* *Search* through code to see where the `else` ends and if there any more code to run afterwards
* If the client is not loaded, locate where the `else` ends and run code inside `else`

My scenario:

* If the client is loaded, invoke the callback with the loaded client and return
* If the client is not loaded, run the rest of the code

test/it/client_it.js
timothyej: Okay, so it does not pass an error as the first argument?

typerandom: It does. But as I said, the done() callback is expected to be called once, and in this case, it might not be.

timothyej: Okay, but that error is handled by the `client.on('error')` then?

typerandom: The error is being thrown and caught by Mocha.

timothyej: Okay, got it :+1:

test/it/helpers.js
typerandom: I'll give you this one.

timothyej: Yaaaaaay :relieved: :beers: :tada:

timothyej:
This needs a rebase before we can merge it :)

typerandom:
It's already done :)

timothyej:
My benchmark (~18% improvement):

**Master**:  
```bash
Execution Time (2015-11-04 13:00:31 UTC)
jshint:test             1.8s  ▇▇▇ 1%
mocha_istanbul:test  2m 4.6s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 98%
Total 2m 7.8s
```

**This branch**:  
```bash
Execution Time (2015-11-04 13:03:12 UTC)
jshint:test              1.6s  ▇▇▇ 2%
mocha_istanbul:test  1m 42.5s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 97%
Total 1m 45.2s
```

timothyej:
Question is; are we okay with losing isolation between tests for this performance improvement?

PR #270, opened 2015-11-03
Fixing oauth error responses

robertjd:
We need to send `invalid_client` for situations where client authentication fails.

### DIFF
--- a/lib/error/ApiAuthRequestError.js
+++ b/lib/error/ApiAuthRequestError.js
@@ -8,9 +8,7 @@ function ApiAuthRequestError(errorData){
   this.name='ApiAuthRequestError';
   this.userMessage = this.message = errorData.userMessage;
   this.statusCode = errorData.statusCode || 400;
-  if(errorData.error){
-    this.error = errorData.error;
-  }
+  this.error = errorData.error;
 }
 utils.inherits(ApiAuthRequestError, Error);
 
--- a/lib/error/ApiAuthRequestError.js
+++ b/lib/error/ApiAuthRequestError.js
@@ -2,12 +2,15 @@
 
 var utils  = require('../utils');
 
-function ApiAuthRequestError(message, statusCode){
+function ApiAuthRequestError(errorData){
   Error.call(this);
   Error.captureStackTrace(this, this.constructor);
   this.name='ApiAuthRequestError';
-  this.userMessage = this.message = message;
-  this.statusCode = statusCode || 400;
+  this.userMessage = this.message = errorData.userMessage;
+  this.statusCode = errorData.statusCode || 400;
+  if(errorData.error){
+    this.error = errorData.error;
+  }
 }
 utils.inherits(ApiAuthRequestError, Error);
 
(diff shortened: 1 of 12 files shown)


lib/error/ApiAuthRequestError.js
```diff
@@ -2,12 +2,15 @@
...
-function ApiAuthRequestError(message, statusCode){
+function ApiAuthRequestError(errorData){
   Error.call(this);
   Error.captureStackTrace(this, this.constructor);
   this.name='ApiAuthRequestError';
-  this.userMessage = this.message = message;
-  this.statusCode = statusCode || 400;
+  this.userMessage = this.message = errorData.userMessage;
+  this.statusCode = errorData.statusCode || 400;
+  if(errorData.error){
```
timothyej: Instead of the `if`, how about:

```javascript
this.error = errorData.error || undefined;
```

Or if `null` and other falsey values are okay to pass on, just assign is directly (without the `if`):

```javascript
this.error = errorData.error;
```

What do you think?

lib/error/ResourceError.js
```diff
@@ -13,6 +13,10 @@ function ResourceError(responseBody) {
   this.message = 'HTTP ' + this.status +
     ', Stormpath ' + this.code + ' (' + this.moreInfo + '): ' +
     this.developerMessage;
+
+  if (responseBody.error) {
```
timothyej: See my other comment.

lib/jwt/jwt-authenticator.js
```diff
@@ -29,7 +29,7 @@ JwtAuthenticator.prototype.withCookie = function withCookie(cookieName){
 };
 
 JwtAuthenticator.prototype.unauthenticated = function unauthenticated(){
-  return new ApiAuthRequestError('Unauthorized', 401);
+  return new ApiAuthRequestError({userMessage:'Unauthorized', statusCode: 401});
```
typerandom: This seems fairly repeated. I'd probably make it an own error that was extended from ApiAuthRequestError. Something like `ApiUnauthorizedError` and have it set the statusCode to `401` and default userMessage to `Unauthorized`. Then add an optional userMessage parameter, so that if it is specified, then it will use that instead. What do you think?

lib/error/ApiAuthRequestError.js
typerandom: I'd go with the second one, `this.error = errorData.error`. There isn't really any reason why anyone would explicitly like to type-check for `undefined` in this case.

lib/jwt/jwt-authenticator.js
robertjd: I agree that we could refactor these error responses to be less dry.  Our eventual goal is to deprecate `application.authenticateApiRequest()` in favor of our new authenticator pattern, i.e. `new Authenticator(application)`, etc.  We can revisit the error strategy when that happens.

lib/error/ApiAuthRequestError.js
robertjd: Thanks. There is no reason to type check for falsey values in this case

lib/jwt/jwt-authenticator.js
typerandom: Okay, awesome!

PR #273, opened 2015-11-06
Feature: Add client getRefreshToken method

typerandom:
Adds a new client method used in order to retrieve a specific refresh token.

### DIFF
--- a/test/it/client_it.js
+++ b/test/it/client_it.js
@@ -44,11 +44,35 @@ describe('Client', function() {
     });
 
     it('should get an access token resource',function(done){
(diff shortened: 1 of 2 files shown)


test/it/client_it.js
```diff
@@ -55,6 +55,30 @@ describe('Client', function() {
...
+
+    before(function(done) {
+      accountCase.before(done);
+    });
+
+    after(function(done) {
+      accountCase.after(done);
+    });
+
+    it('should get an refresh token resource',function(done){
```
timothyej: Typo: `an` => `a`

PR #279, opened 2015-11-11
Fix: local validation needs to apply the account href

robertjd:


rdegges:
might wanna move that package.json patch into a new PR. other than that, +1

PR #293, opened 2015-11-20
Removing remote strategy loading

robertjd:
@typerandom can you sanity check this for me?

`stormpath-config@0.0.13` affects master in this repo, in that you cannot create a client without specifying application configuration.  This is not what we want from the Node SDK, as you don't always want to be working with an application - that requirement is delegated to the framework integration (which does require an application).

Here is the related change in the express library:

https://github.com/stormpath/express-stormpath/commit/3df9ab26afc815550a709f33dcad329b7380bd41

### DIFF
--- a/lib/configLoader.js
+++ b/lib/configLoader.js
@@ -1,22 +1,11 @@
 var path = require('path');
 
-var Client = require('./Client');
 var stormpathConfig = require('stormpath-config');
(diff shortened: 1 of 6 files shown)


typerandom:
It's awesome if we can move this from the SDK to the integration. It makes a lot more sense there.

And the changes looks great, so I approve! :beers:

PR #318, opened 2015-12-21
Fix: Cache eviction of customData when calling account.save()

typerandom:
Invalidate customData when setting customData on an account and then calling `account.save()`.

Previously the customData resource would be updated, but the cache would still be stale.

Closes https://github.com/stormpath/express-stormpath/issues/156

### DIFF
--- a/lib/resource/Account.js
+++ b/lib/resource/Account.js
@@ -100,8 +100,8 @@ Account.prototype.save = function saveAccount(){
   // If customData, then inject our own callback and invalidate the
   // customData resource cache when the account finishes saving.
   if (self.customData) {
-    var originalCallback = args.length > 0 && typeof args[args.length - 1] === 'function' ?
-      args.pop() : function nop() {};
+    var lastArg = args.length > 0 ? args[args.length - 1] : null;
+    var originalCallback = typeof lastArg === 'function' ? args.pop() : function nop() {};
 
     args.push(function newCallback() {
       self.dataStore._evict(self.customData.href, originalCallback);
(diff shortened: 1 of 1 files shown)


lib/resource/Account.js
```diff
@@ -94,7 +94,20 @@ Account.prototype.getApiKeys = function getApiKeys(options,callback) {
...
 Account.prototype.save = function saveAccount(){
   var self = this;
-  var args = arguments;
+
+  var args = Array.prototype.slice.call(arguments);
+
+  // If customData, then inject our own callback and invalidate the
+  // customData resource cache when the account finishes saving.
+  if (self.customData) {
+    var originalCallback = args.length > 0 && typeof args[args.length - 1] === 'function' ?
```
timothyej: Crazy oneliner bro! How about this instead:

```javascript
var lastArg = args.length > 0 ? args[args.length - 1] : null;
var originalCallback = typeof lastArg === 'function' ? args.pop() : function noop() {};
```

(Typo: `nop` => `noop`)

typerandom: Yeah, that definitely looks better! :) :+1: 

Also regarding NOP, it's not a typo: https://en.wikipedia.org/wiki/NOP

robertjd:
I'm happy to merge this as is, but I wonder how else this could be done so that it will fix this issue for all resources that have custom data?

typerandom:
@robertjd Yes, definitely. A hack would be to do it in `InstanceResource.save()` and have a check if the instance has a `customData` attribute. But that isn't pretty.

A better approach would be for `InstanceResource.save()` to check if there are child entities have been modified, and if they have, then save them before saving itself. But it's possible that it could have side effects that I'm unaware of.

PR #320, opened 2015-12-21
Refactor: Remove redundant constructor return

typerandom:
Remove redundant 'return this' used in constructors.


timothyej:
Very nice! :+1:

PR #321, opened 2015-12-22
Refactor: Remove redundant dynamic args mapping

typerandom:
Replaces manual dynamic argument mapping with a utility function.

I.e.

```javascript
var args = Array.prototype.slice.call(arguments);
var href = args.shift();
var callback = args.pop();
var options = (args.length > 0) ? args.shift() : null;
```

Can be replaced with:

```javascript
var args = utils.resolveArgs(arguments, ['href', 'options', 'callback']);
```

**Note:**

This is just the first step in refactoring the dynamic argument mapping. The second will be to build out the `resolveArgs()` method so that one can write, e.g.: `['href!', 'options', 'callback!']`. I.e. giving the ability to require an argument to be present, and an error to be thrown if it isn't. Also removing the need for the `rightToLeft` parameter.

### DIFF
--- a/lib/resource/Group.js
+++ b/lib/resource/Group.js
@@ -6,37 +6,29 @@ var utils = require('../utils');
 function Group() {
   Group.super_.apply(this, arguments);
 }
+
 utils.inherits(Group, DirectoryChildResource);
 
 Group.prototype.addAccount = function addGroupAccount(/* accountOrAccountHref, [options,] callback */) {
-  var args = Array.prototype.slice.call(arguments);
-  var account = args.shift();
-  var callback = args.pop();
-  var options = (args.length > 0) ? args.shift() : null;
+  var args = utils.resolveArgs(arguments, ['account', 'options', 'callback']);
 
-  if (typeof account === 'string') {
-    account = {
-      href: account
+  if (typeof args.account === 'string') {
+    args.account = {
+      href: args.account
     };
   }
 
-  return this._createGroupMembership(account, this, options, callback);
+  return this._createGroupMembership(args.account, this, args.options, args.callback);
 };
 
 Group.prototype.getAccounts = function getGroupAccounts(/* [options,] callback */) {
-  var args = Array.prototype.slice.call(arguments);
-  var callback = args.pop();
-  var options = (args.length > 0) ? args.shift() : null;
-
-  return this.dataStore.getResource(this.accounts.href, options, require('./Account'), callback);
+  var args = utils.resolveArgs(arguments, ['options', 'callback'], true);
+  return this.dataStore.getResource(this.accounts.href, args.options, require('./Account'), args.callback);
 };
 
 Group.prototype.getAccountMemberships = function getGroupAccountMemberships(/* [options,] callback */) {
-  var args = Array.prototype.slice.call(arguments);
-  var callback = args.pop();
-  var options = (args.length > 0) ? args.shift() : null;
-
-  return this.dataStore.getResource(this.accountMemberships.href, options, require('./GroupMembership'), callback);
+  var args = utils.resolveArgs(arguments, ['options', 'callback'], true);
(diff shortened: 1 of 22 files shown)


lib/utils.js
```diff
@@ -19,6 +19,35 @@ function shallowCopy(src, dest) {
...
+  var takeFromRight = rightToLeft;
+  var args = Array.prototype.slice.call(argumentsObject);
+
+  while (nameMap.length) {
+    var value = null;
+    var name = take(nameMap, takeFromRight);
+
+    if (args.length > 0) {
+      value = take(args, takeFromRight);
+      takeFromRight = !takeFromRight;
```
robertjd: What is this for?

typerandom: Timothy asked the same question. You can see my answer here:

https://github.com/stormpath/stormpath-sdk-node/commit/4e24e1e89b4e1900d89fbe6c54cc31e2f8261a66#diff-50e3aa130a4f97a42ee2cf111c7b1d9dR42

robertjd: I'm skeptical that this is going to work for all situations.  I'd want to look at the test coverage for every public function that uses this.  I just linked the current master of `express-stormpath` to this branch of `stormpath`, and I get a handful of failures that I don't see with the latest stable release of `stormpath`.  Can you try as well?

typerandom: It's a hack. But I'm pretty certain that it should work since it covers the two most common mapping scenarios: `href, [options,] callback` and `[options,] callback`.

And as I wrote before, it's a hack and I have an idea of how to fix it long-term :) But I haven't tried using the `express-stormpath` module and linking this in, so will do that and see what it throws.

typerandom: @robertjd What errors were you seeing? I'm actually seeing errors right now when using both the SDK master and express master. Are you seeing the same?

typerandom: @robertjd After I applied the fix in https://github.com/stormpath/stormpath-sdk-node/pull/323 (fix of typo from previous refactor) and ran this branch with the `express-stormpath` test suite, it all passed successfully.

Could you try the same?

robertjd: Yes, #323 appears to fix the issue.  Want to rebase this branch against master?

typerandom: It's rebased now.

robertjd:
I can confirm that the most recent reabse will work successfully with the express-stormpath.  @typerandom is currently investigate an issue with our test runner (not all tests are being run), so we will wait until those issues are resolved so that we can run this module's suite against the fixes that may be coming to the rest runner.

timothyej:
@typerandom Did you solve the test problem?

typerandom:
As you know, it's not been solved yet.

timothyej:
Working on it as we speak :)

typerandom:
Yaayyy :)

robertjd:
This is lingering in the review column, but we need to resolve #334 first.

lib/resource/Group.js
```diff
@@ -6,37 +6,29 @@ var utils = require('../utils');
...
 }
+
 utils.inherits(Group, DirectoryChildResource);
 
 Group.prototype.addAccount = function addGroupAccount(/* accountOrAccountHref, [options,] callback */) {
-  var args = Array.prototype.slice.call(arguments);
-  var account = args.shift();
-  var callback = args.pop();
-  var options = (args.length > 0) ? args.shift() : null;
+  var args = utils.resolveArgs(arguments, ['account', 'options', 'callback']);
```
timothyej: This scenario doesn't seem to work. The `options` argument should be optional but it isn't, so calling `.addAccount()` without `options` will result in an error.

typerandom:
@timothyej @robertjd This has been fixed (except for the lowered coverage). Please review.

timothyej:
I've reviewed this and everything looks and works fine 👍But please fix the two bugs I mentioned and then we can merge this :)

timothyej:
I've fixed the bug and I've merged with `master`. At the same time I implemented the new args mapping on the new `Organization` methods.

PR #328, opened 2015-12-27
Fix: Improve request executor error message

### LINKED ISSUE #160: Request executor error message contains the string "[object Object]"
Using node 0.12 on Ubuntu, I've seen error messages like this:

```
Unable to execute http request [object Object]: getaddrinfo ENOTFOUND api.stormpath.com
```

Originating here: https://github.com/stormpath/stormpath-sdk-node/blob/master/lib/ds/RequestExecutor.js#L90

It's a minor issue, but it might help debugging if it included more specific info about the `req` object, such as the method and URL.

typerandom:
Adds a more detailed error message for the RequestExecutor.

Fixes #160

### DIFF
--- a/lib/ds/RequestExecutor.js
+++ b/lib/ds/RequestExecutor.js
@@ -92,6 +92,7 @@ RequestExecutor.prototype.execute = function executeRequest(req, callback) {
     if (err) {
       var wrapper = new Error('Unable to execute http request ' + req.method + ' ' + req.uri + ': ' + err.message);
       wrapper.inner = err;
+      wrapper.stack = '';
       return callback(wrapper, null);
     }
 
--- a/lib/ds/RequestExecutor.js
+++ b/lib/ds/RequestExecutor.js
@@ -68,22 +68,23 @@ RequestExecutor.prototype.execute = function executeRequest(req, callback) {
     throw new Error('request.uri field is required.');
   }
 
-  //don't override the defaults: ensure that the options arg is request-specific:
+  // Don't override the defaults: ensure that the options arg is request-specific.
   var options = utils.shallowCopy(this.options, {});
 
-  if (req.method) { //defaults to GET
-    options.method = req.method;
-  }
+  req.method = req.method || 'GET';
+
+  options.method = req.method;
   options.uri = this.qualify(req.uri);
+
   if (req.query) {
     options.qs = req.query;
   }
+
   if (req.body && req.body.form){
     options.form = req.body.form;
-  }
-  else if (req.body) {
+  } else if (req.body) {
     options.body = req.body;
-    options.json = true; //all Stormpath resources are JSON
+    options.json = true; // All Stormpath resources are JSON
   }
 
   this.requestAuthenticator.authenticate(options);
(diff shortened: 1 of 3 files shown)


lib/ds/RequestExecutor.js
```diff
@@ -90,8 +90,9 @@ RequestExecutor.prototype.execute = function executeRequest(req, callback) {
 
   request(options, function onRequestResult(err, response, body) {
     if (err) {
-      var wrapper = new Error('Unable to execute http request ' + req + ': ' + err.message);
+      var wrapper = new Error('Unable to execute http request ' + req.method + ' ' + req.uri + ': ' + err.message);
```
robertjd: req.method will be undefined if this is a GET request that has not explicitly declared a method to be used.

lib/ds/RequestExecutor.js
```diff
@@ -90,8 +90,9 @@ RequestExecutor.prototype.execute = function executeRequest(req, callback) {
...
       wrapper.inner = err;
+      wrapper.stack = '';
```
robertjd: Why erase the stack?

test/sp.ds.requestExecutor_test.js
```diff
@@ -107,7 +107,33 @@ describe('ds:', function () {
...
+      it('should include HTTP details in message in case of request error', function (done) {
+        // This triggers one of the possible http request errors
+        var uri = 'http://doesntexist/v1/test';
+
+        var cbSpy = sinon.spy(function (err, body) {
+          err.should.be.an.instanceof(Error);
+
+          err.should.have.property('stack')
+            .that.is.not.null.and
+            .that.is.not.undefined;
```
robertjd: Is this test running?  In the following example, the stack is null:

```

var applicationHref = 'https://api2.stormpath.com/blah';

client.createResource(applicationHref,function (err) {
  console.log(err.stack);
});
```

typerandom: It should be, but maybe there's some other error that is being thrown?

lib/ds/RequestExecutor.js
typerandom: This was actually not erasing, but more setting a default stack. Since mocha threw an error of it was undefined. So I just assumed the stack was being set when you threw an error. But maybe it's being set when you create it... Anyhow, it's to fix a Mocha issue. Without this Mocha will just crash.

lib/ds/RequestExecutor.js
typerandom: Okay. Then I'll just fix the `req.method`, so that if it is undefined, then it will default to `GET`.

timothyej:
Any updates on this one?

timothyej:
I've refactored the tests for this (273e9d22a7c57a69420974029a4889f0ac3adfb7):

**Before:**  
```bash
✓ should include the original request error in case of request error
✓ should include HTTP details in message in case of request error
```

**After:**  
```bash
in case of request error
  when request method is GET
    body
      ✓ should be null
    error
      ✓ should be an instance of Error
      stack property
        ✓ should not be empty/null/undefined
      inner property
        ✓ should be an instance of Error
        ✓ should have the "code" property set to "ENOTFOUND"
      message property
        ✓ should include request method
        ✓ should include request uri
        ✓ should include error.inner.code
  when request method is undefined
    error
      message property
        ✓ should include request method as "GET"
```

timothyej:
:+1:

PR #329, opened 2015-12-27
Fix: Correctly pass TTL to authentication result

### LINKED ISSUE #324: Can't set expiration on JWT
AuthenticationResult.getAccessTokenResponse() will always return an undefined "expires_in" property, because nJwt does not use a ttl property as stormpath-sdk_node expects. nJwt uses "exp" instead of "ttl".

typerandom:
Fixes so that TTL is correctly passed to from authenticators to Authentication Result and so that the access token response correctly translates the JWT `exp` into the access token response `expires_in`.

Fixes #324

### DIFF
--- a/lib/authc/OAuthBasicExchangeAuthenticator.js
+++ b/lib/authc/OAuthBasicExchangeAuthenticator.js
@@ -47,7 +47,7 @@ OAuthBasicExchangeAuthenticator.prototype.authenticate = function authenticate(c
 
         result.forApiKey = apiKey;
         result.application = self.application;
-        result.expiration = self.ttl;
+        result.ttl = self.ttl;
 
         self.buildTokenResponse(apiKey, function onTokenResponse(err, tokenResponse) {
           if (err) {
--- a/lib/authc/OAuthBasicExchangeAuthenticator.js
+++ b/lib/authc/OAuthBasicExchangeAuthenticator.js
@@ -47,6 +47,7 @@ OAuthBasicExchangeAuthenticator.prototype.authenticate = function authenticate(c
 
         result.forApiKey = apiKey;
         result.application = self.application;
+        result.expiration = self.ttl;
 
         self.buildTokenResponse(apiKey, function onTokenResponse(err, tokenResponse) {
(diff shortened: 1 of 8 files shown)


lib/resource/AuthenticationResult.js
```diff
@@ -34,16 +44,18 @@ AuthenticationResult.prototype.getAccessToken = function getAccessToken(jwt) {
...
 
 AuthenticationResult.prototype.getAccessTokenResponse = function getAccessTokenResponse(jwt) {
-
   jwt = jwt || this.getJwt();
+
   var resp = {
     'access_token': jwt.compact(),
     'token_type': 'Bearer',
-    'expires_in': jwt.ttl
+    'expires_in': jwt.body.exp
```
robertjd: The `expires_in` field is not an absolute date, it is the lifetime of the token in seconds.  As such, this should be the original `ttl` value that was provided to `authenticateApiRequest`.  See https://tools.ietf.org/html/rfc6749#section-5.1

lib/authc/OAuthBasicExchangeAuthenticator.js
```diff
@@ -47,6 +47,7 @@ OAuthBasicExchangeAuthenticator.prototype.authenticate = function authenticate(c
 
         result.forApiKey = apiKey;
         result.application = self.application;
+        result.expiration = self.ttl;
```
robertjd: I'm not so sure that this should be renamed to `expiration`, see my other comment on this PR.

lib/resource/AuthenticationResult.js
typerandom: Yeah, I see how this is wrong now. Thanks for pointing that out! :+1:

lib/authc/OAuthBasicExchangeAuthenticator.js
typerandom: As pointed out in your other comment, I also see how this should be renamed to `ttl` instead.

timothyej:
LGTM :+1:

PR #330, opened 2015-12-27
Fix: Remove refresh_token option validation in refresh grant authenticator

### LINKED ISSUE #306: Incorrectly throwing an error when request data does not have refresh_token property
The refresh_token oauth authenticator is not returning the correct error (invalid_request with statusCode 400) when no refresh_token is specified on the request body.
The authenticator should be returning the error that came from Stormpath API and instead is throwing an error that hides this error returned by the Stormpath API.

The verification on line 31, if the refresh_token field is present on request data and the consequent throw on file /lib/oauth/refresh-grant.js, is the incorrect code.
```javascript
  }else if (!data.refresh_token){
    throw new Error('data does not have refresh_token property');
  }
```

typerandom:
For consistency, remove validation of the `refresh_grant` option in the refresh token grant authenticator. I.e. if missing, then let the error fall through and be handled by the API.

Fixes #306

### DIFF
--- a/lib/oauth/refresh-grant.js
+++ b/lib/oauth/refresh-grant.js
@@ -27,8 +27,6 @@ OAuthRefreshTokenGrantAuthenticator.prototype.authenticate = function authentica
   var application = this.application;
   if(arguments.length !==2 ){
     throw new Error('Must call authenticate with (data,callback)');
-  }else if (!data.refresh_token){
-    throw new Error('data does not have refresh_token property');
   }else{
     var href = application.href + '/oauth/token';
     var formData = {


typerandom:
Currently coveralls is reporting a decrease of 8.4% in coverage. Makes no sense at all since it's complexity that has been removed, and it's only two lines...

robertjd:
@typerandom perhaps this is related to the problems we were discussing his morning RE the `Resources:` tests not running?

I've reviewed the code change and +1 it, but let's figure out the test issue before merging

timothyej:
This PR is getting old, and works fine for me, so I'm merging this now :+1:

PR #340, opened 2016-01-06
Feature: Add organization resource methods

### LINKED ISSUE #245: Implement methods on Organization resource
Ned to implement the following methods on the Organization resource:

createAccount()

getAccounts()

getCustomData()

getDefaultAccountStore()

getDefaultGroupStore()

getGroups()

getIdSiteModel()

typerandom:
Adds the following methods to the Organization resource:

* createAccount()
* getAccounts()
* getCustomData()
* getDefaultAccountStore()
* getDefaultGroupStore()
* getGroups()
* getIdSiteModel()

Fixes #245.


rdegges:
+1 looks great to me, go for merge!

lib/resource/Organization.js
```diff
@@ -11,8 +14,81 @@ function Organization() {
...
+  var args = Array.prototype.slice.call(arguments);
+
+  var callback = args.pop();
+  var options = (args.length > 0) ? args.shift() : null;
+
+  if (!this.defaultGroupStoreMapping) {
+    return callback();
+  }
+
+  return this.dataStore.getResource(this.defaultGroupStoreMapping.href, options, OrganizationAccountStoreMapping, callback);
```
timothyej: Is it supposed to be `OrganizationAccountStoreMapping` here?

typerandom: Yes, this is as intended.

test/sp.resource.organization_test.js
```diff
@@ -0,0 +1,280 @@
...
+var Organization = require('../lib/resource/Organization');
+var OrganizationAccountStoreMapping = require('../lib/resource/OrganizationAccountStoreMapping');
+
+/*jshint -W030 */
+describe('Resources: ', function () {
+  describe('Organization resource', function () {
+    var dataStore = new DataStore({client: {apiKey: {id: 1, secret: 2}}});
+
+    describe('create account', function () {
+      describe('if accounts are set', function () {
```
robertjd: This describe should be removed, it's describing the setup of the test mock (not the code under test).

robertjd: I feel that this part suite needs to be restructure liked this, so that it describes the features and functions under test:

```
create account
  if called without options
    it should post the account
  if called with options
    it should post the account with options
```

timothyej: I'm looking into this :+1:

timothyej:
I just rewrite the unit tests for `resource/Organization.js`. At the same time I found a bug when parsing arguments for `createAccountStoreMapping`: 175ce35a536d0dbd43a3150968469e486dd6015c

**Before:**  
```bash
Resources:
  Organization resource
    create account
      if accounts are set
        ✓ should create account
    get accounts
      if accounts are set
        ✓ should get accounts
    get custom data
      if customData are set
        ✓ should get accounts
    get default account store
      if defaultAccountStoreMapping are set
        ✓ should get accounts
    get default group store
      if defaultGroupStoreMapping are set
        ✓ should get accounts
    get groups
      if groups href are set
        ✓ should get groups
    get id site model
      if idSiteModel are set
        ✓ should get accounts
```

**After:**  
```bash
resource/Organization.js
  Organization
    ✓ should inherit from InstanceResource
    constructor
      ✓ should call super_ with the same arguments
    prototype
      createAccount method
        when called as createAccount(account, options, callback)
          ✓ should call dataStore.createResource with accounts.href, options, account, Account, callback
        when called as createAccount(account, callback)
          ✓ should call dataStore.createResource with accounts.href, null, account, Account, callback
      getAccounts method
        when called as getAccounts(options, callback)
          ✓ should call dataStore.getResource with accounts.href, options, Account, callback
        when called as getAccounts(callback)
          ✓ should call dataStore.getResource with accounts.href, null, Account, callback
      getCustomData method
        when called as getCustomData(options, callback)
          ✓ should call dataStore.getResource with customData.href, options, CustomData, callback
        when called as getCustomData(callback)
          ✓ should call dataStore.getResource with customData.href, null, CustomData, callback
      getDefaultAccountStore method
        when called as getDefaultAccountStore(options, callback)
          ✓ should call dataStore.getResource with defaultAccountStoreMapping.href, options, OrganizationAccountStoreMapping, callback
          when defaultAccountStoreMapping is undefined
            ✓ should invoke the callback
            ✓ should not invoke dataStore.getResource
        when called as getDefaultAccountStore(callback)
          ✓ should call dataStore.getResource with defaultAccountStoreMapping.href, null, OrganizationAccountStoreMapping, callback
          when defaultAccountStoreMapping is undefined
            ✓ should invoke the callback
            ✓ should not invoke dataStore.getResource
      getDefaultGroupStore method
        when called as getDefaultGroupStore(options, callback)
          ✓ should call dataStore.getResource with defaultGroupStoreMapping.href, options, OrganizationAccountStoreMapping, callback
          when getDefaultGroupStore is undefined
            ✓ should invoke the callback
            ✓ should not invoke dataStore.getResource
        when called as getDefaultGroupStore(callback)
          ✓ should call dataStore.getResource with defaultGroupStoreMapping.href, null, OrganizationAccountStoreMapping, callback
          when defaultGroupStoreMapping is undefined
            ✓ should invoke the callback
            ✓ should not invoke dataStore.getResource
      getGroups method
        when called as getGroups(options, callback)
          ✓ should call dataStore.getResource with groups.href, options, Group, callback
        when called as getGroups(callback)
          ✓ should call dataStore.getResource with groups.href, null, Group, callback
      getIdSiteModel method
        when called as getIdSiteModel(options, callback)
          ✓ should call dataStore.getResource with idSiteModel.href, options, IdSiteModel, callback
        when called as getIdSiteModel(callback)
          ✓ should call dataStore.getResource with idSiteModel.href, null, IdSiteModel, callback
      getAccountStoreMappings method
        when called as getAccountStoreMappings(options, callback)
          ✓ should call dataStore.getResource with accountStoreMappings.href, options, OrganizationAccountStoreMapping, callback
        when called as getAccountStoreMappings(callback)
          ✓ should call dataStore.getResource with accountStoreMappings.href, null, OrganizationAccountStoreMapping, callback
      createAccountStoreMapping method
        when called as createAccountStoreMapping(mapping, options, callback)
          ✓ should call dataStore.createResource with "/organizationAccountStoreMappings", options, mapping, OrganizationAccountStoreMapping, callback
          ✓ should add the organization to the new mapping object
        when called as createAccountStoreMapping(mapping, callback)
          ✓ should call dataStore.createResource with "/organizationAccountStoreMappings", null, mapping, OrganizationAccountStoreMapping, callback
          ✓ should add the organization to the new mapping object
      createAccountStoreMappings method
        when called as createAccountStoreMappings(mappings, callback)
          ✓ should call createAccountStoreMapping with each mapping
          ✓ should invoke callback after it has created each account store mapping
```

robertjd:
Hey @timothyej , thank you for the work on the tests, the structure is looking better!  However I'm going to require some more improvements.  Please look at [stormpath_assertion_authenticator_it.js](https://github.com/stormpath/stormpath-sdk-node/blob/master/test/it/stormpath_assertion_authenticator_it.js) for a reference of how I'd like to see our tests being written.

Those tests have the following features:

* The test descriptors describe what is being tested, and what the expected behavior (output) is.

* The test code minimizes the need for spies and spy assertions, it does this by using more mocks and stubs.

Taking this part of the suite as an example:

```
getIdSiteModel method
  when called as getIdSiteModel(options, callback)
    ✓ should call dataStore.getResource with idSiteModel.href, options, IdSiteModel, callback
  when called as getIdSiteModel(callback)
    ✓ should call dataStore.getResource with idSiteModel.href, null, IdSiteModel, callback
```

The issue I see is that the statements don’t describe what I should get back from this operation, they are describing an internal working of the `getIdSiteModel()` method.  I don’t always care about those internal workings, I care more about the inputs and outputs of the interface that is under test (`getIdSiteModel()`).

Given these requirements, I would expect the test output to read more like this:

```
Organization
  .getIdSiteModel(callback)
    ✓ should return the organization’s IdSiteModel
  .getIdSiteModel(options, callback)
    ✓ should pass the options to the request executor
    ✓ should return the organization’s IdSiteModel
```

Thanks again, looking forward to the next review! I’m really excited about the improvements that you’ve been making to the tests on this library <3

timothyej:
Thanks for the great feedback, @robertjd :) I totally agree with you, and I was a bit sceptical myself on how to actually test these methods, as all they do is to *parse* the arguments and passes them along to the data store. And as we saw, it's pretty easy to introduce bugs into this logic. But with your suggestions, we test that the arguments we send in get passed along as intended.

timothyej:
As I agree with you that it's the interface that should be tested, these kind of methods are a bit tricky to test as they just act as proxies.

Regarding your example with the `.getIdSideModel()` test, I can't test that the method actually returns the organization's ID Site model, as that logic relies inside the `dataStore.getResource()` method. And that logic depends on that this method calls it properly.

So the actual logic here is to *parse* the arguments and pass them along to the data store, so I decided to test that.

If we want to test that the `.getIdSideModel()` method (and all the other methods) actually returns the organization's ID Site model, we should create an integration test that tests this method together with the `dataStore.getResource()` method, which actually someone has already started doing [here](https://github.com/stormpath/stormpath-sdk-node/blob/master/test/it/organization_it.js).

Let me know what you think!

**The result:**  
```bash
resource/Organization.js
  Organization
    ✓ should inherit from InstanceResource
    constructor
      ✓ should call super_ with the same arguments
    .createAccount(account, options, callback)
      ✓ should pass the options to dataStore.createResource
      ✓ should pass the callback to dataStore.createResource
      ✓ should return the value from dataStore.createResource
    .createAccount(account, callback)
      ✓ should pass the callback to dataStore.createResource
      ✓ should return the value from dataStore.createResource
    .getAccounts(options, callback)
      ✓ should pass the options to dataStore.getResource
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
    .getAccounts(callback)
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
    .getCustomData(options, callback)
      ✓ should pass the options to dataStore.getResource
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
    .getCustomData(callback)
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
    .getDefaultAccountStore(options, callback)
      ✓ should pass the options to dataStore.getResource
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
      when defaultAccountStoreMapping is undefined
        ✓ should invoke the callback
        ✓ should not invoke dataStore.getResource
        ✓ should return the value from the callback
    .getDefaultAccountStore(callback)
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
      when defaultAccountStoreMapping is undefined
        ✓ should invoke the callback
        ✓ should not invoke dataStore.getResource
        ✓ should return the value from the callback
    .getDefaultGroupStore(options, callback)
      ✓ should pass the options to dataStore.getResource
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
      when getDefaultGroupStore is undefined
        ✓ should invoke the callback
        ✓ should not invoke dataStore.getResource
        ✓ should return the value from the callback
    .getDefaultGroupStore(callback)
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
      when getDefaultGroupStore is undefined
        ✓ should invoke the callback
        ✓ should not invoke dataStore.getResource
        ✓ should return the value from the callback
    .getGroups(options, callback)
      ✓ should pass the options to dataStore.getResource
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
    .getGroups(callback)
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
    .getIdSiteModel(options, callback)
      ✓ should pass the options to dataStore.getResource
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
    .getIdSiteModel(callback)
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
    .getAccountStoreMappings(options, callback)
      ✓ should pass the options to dataStore.getResource
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
    .getAccountStoreMappings(callback)
      ✓ should pass the callback to dataStore.getResource
      ✓ should return the value from dataStore.getResource
    .createAccountStoreMapping(mapping, options, callback)
      ✓ should pass the options to dataStore.createResource
      ✓ should pass the callback to dataStore.createResource
      ✓ should add the organization to the new mapping object
      ✓ should return the value from dataStore.createResource
    .createAccountStoreMapping(mapping, callback)
      ✓ should pass the callback to dataStore.createResource
      ✓ should add the organization to the new mapping object
      ✓ should return the value from dataStore.createResource
    .createAccountStoreMappings(mappings, callback)
      ✓ should call createAccountStoreMapping with each mapping
      ✓ should invoke callback after it has created each account store mapping
      ✓ should return undefined
```

robertjd:
Thanks for the updates!  This tests look a lot cleaner and I don’t see the need to make any more changes to the tests.  Please read down for more thoughts on tests.

However I just noticed that `Organization.getDefaultAccountStore()` is not returning an account store, it is retuning an account store mapping.  Thus this function is named incorrectly.  Unfortunately we have this same problem with `Application.getDefaultAccountStore()`, and that function has been in master forever.  I’ll open a separate issue for that.

Alas, we need to fix this PR such that:

* `Organization.getDefaultAccountStore()` and `Organization.getDefaultGroupStore()` will return the account store that is identified by the mapping.  When doing the first step of fetching the mapping, use the option to expand the account store.  This will get that resource cached locally, so that when you make the next request to get the account store, it will already be cached.

* `Organization.getDefaultAccountStoreMapping()` and `Organization.getDefaultGroupStoreMapping()` need to be implemented and will return the account store mapping.

My bad for not seeing this earlier.

Okay, now.. big thought on tests:  Structurally I still see problems with testing the data store interaction via this interface, but this is not a problem that we should spend more time on with this PR.  I would prefer that we use a mock request executor that can respond with contextual mocks.  Thus you test the input (options, or no options) and the output (a response that happens with options, and one that happens without options).  The datastore code will be transparently covered.  Looking forward, I believe we need to get rid of “unit” vs “integration” tests.  Instead we should have one test suite with two different fixtures: one that provides mocks (via a mock request executor) and one that hits the production API.  The only real reason for the two is speed: local mocks will be way faster than working against the production API.

timothyej:
@robertjd Interesting thoughts, I would like to hear more about them later :) And as I can see that you created an issue (#376) regarding `getDefaultAccountStore()` and `getDefaultGroupStore()`, I will now merge this PR 👍

Issue #342, opened 2016-01-08
OauthAccessTokenAuthenticator does not attach application to AuthenticationResult


areichman:
When creating JWTs using the `AuthenticationResult.getJwt` method, it is expected that the `AuthenticationResult` object has a reference to an `application` property so it can properly set the `iss` value in the token.

This `application` object is set by the `OAuthBasicExchangeAuthenticator` when creating an `AuthenticationResult`:

```javascript
var result = new AuthenticationResult(apiKey,self.application.dataStore);
result.forApiKey = apiKey;
result.application = self.application;
// ...
callback(null, result);
```

But it is not set by the `OauthAccessTokenAuthenticator`:

```javascript
callback(null,new AuthenticationResult(data,self.application.dataStore));
```

The result is that if you try to authenticate a request using an existing token (in the header, body, or url), you cannot call the `getJwt` method to create a new token since that method will fail trying to lookup the `application.href` value.

timothyej:
Thanks for reporting and fixing this! I've issued a new PR based on your changes: #343

robertjd:
Hi Aaron, thanks for the report!

The `getJwt` method should only available on an `AuthenticationResult` instance if the authentication result is created during the `grant_type=client_credentials` request, where the client is presenting api key credentials in exchange for an access token.  It’s presence otherwise is a mistake on our part, my apologies.  

Can you tell us more about your use case?  Typically, with the Oauth2 scheme, the client has to present a refresh token in order to get a new access token.  Access tokens are not used to get new access tokens.

We do support access tokens and refresh tokens for our password grant flow.  Are you working with our client_crendential flow, or passwords?

Thanks in advance.

areichman:
Hi Robert-
In our API, all of our token access is off a `/oauth/token` endpoint. From there, we primarily support `grant_type=client_credentials` for our end users. But for our own first class Backbone app, we also have undocumented `password` and `refresh_token` support as well. This is the basic setup for our API method:

```javascript
// user sends api keys
if (grant_type == 'client_credentials') {
  application.authenticateApiRequest({request: req}, function(err, authResult) {
    handleAuthResult(authResult);
  });
} 

// user sends username and password
else if (grant_type == 'password') {
  application.authenticateAccount({username: req.body.username, password: req.body.password}, function(err, authResult) {
    handleAuthResult(authResult);
  });
}

// user sends token
else if (grant_type == 'refresh_token') {
[...]

PR #343, opened 2016-01-11
Add application and ttl references to the authentication result

timothyej:
Thanks to @areichman for reporting and fixing #342.

This PR adds the `application` and `ttl` as a reference to the returned `AuthenticationResult`.

@typerandom please review.

### DIFF
--- a/lib/authc/OauthAccessTokenAuthenticator.js
+++ b/lib/authc/OauthAccessTokenAuthenticator.js
@@ -85,6 +85,7 @@ OauthAccessTokenAuthenticator.prototype.authenticate = function authenticate(cal
 
         var result = new AuthenticationResult(data,self.application.dataStore);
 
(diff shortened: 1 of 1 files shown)


typerandom:
Why are `ttl` and `application` added as properties and not in the constructor of `AuthenticationResult`?

timothyej:
I just replicated how it was done on line 89. We could of course refactor that and make it as input arguments in the constructor?

typerandom:
If it doesn't break how `AuthenticatedResult` is used then I'd suggest that.

robertjd:
Thanks @timothyej for taking this one.   However, I'm going to declare this as wont-fix.  We shouldn't be allowing this flow to begin with, this was a mistake with the original implementation.  Instead we need to finish implementing the new authenticators pattern, as defined by #357

PR #345, opened 2016-01-12
Feature: SAML Support

### LINKED ISSUE #332: SAML support
In order to work with SAML via our REST API, the following needs to be done in this SDK.

* Create a `SamlProvider` class which has the methods `getAttributeStatementMappingRules` and `getServiceProviderMetadata`.  The `Directory.getProvider` method will need to use this new class if the provider resource has a provider ID of `saml`, otherwise it can continue using the generic `Provider` class.

* Create a `SamlIdpUrlBuilder`, which will be exposed as `require(‘stormpath’).SamlIdpUrlBuilder`.  The usage will be `var samlIdpUrlBuilder = new SamlIdpUrlBuilder(application)`.  Then the builder will have the method `samlIdpUrlBuilder.build()`, which will return a string which is the location that the user should be redirected to.  To get this URL , the builder will need to implement [this flow](https://docs.stormpath.com/rest/product-guide/latest/005_auth_n.html#step-1-initiate-the-flow)

* Create the a new authenticator,  `StormpathAssertionAuthenticator` that will implement [this part](https://docs.stormpath.com/rest/product-guide/latest/005_auth_n.html#step-5-stormpath-response-with-jwt) of the flow (validating the incoming JWT response and returning an authentication result, the result should have a `getAccount` method).  This authenticator will be used for handling SAML callbacks as well as ID Site callbacks (we'll deprecate `Application.handleIdSiteResult` in favor of this new authenticator.

typerandom:
Adds the ability to authenticate SAML requests.

Fixes #332.


lib/authc/AssertionAuthenticationResult.js
```diff
@@ -0,0 +1,37 @@
...
+var nJwt = require('njwt');
+var Account = require('../resource/Account');
+
+function AssertionAuthenticationResult(dataStore, data) {
+  this.dataStore = dataStore;
+
+  // Copy data properties.
+  if (data) {
+    for (var key in data) {
+      if (this[key]) {
```
timothyej: This key will still be copied if the value of `this[key]` is [falsey](http://stackoverflow.com/a/19839953/1685939), e.g. `0` or `false`. Is this intended or should it only apply for `undefined`/`null`/non-existent values?

lib/resource/Directory.js
```diff
@@ -75,7 +77,19 @@ Directory.prototype.getProvider = function getProvider(/* [options,] callback */
     return callback();
   }
 
-  return this.dataStore.getResource(this.provider.href, options, require('./Provider'), callback);
+  function providerFactory(provider) {
```
timothyej: This doesn't seem to work properly, it gives me an object of type `InstanceResource`, and not `SamlProvider`:

```javascript
InstanceResource {
  href: 'http://localhost:7001/v1/directories/7MOtq5XXpQVqnD7WgpxoW4/provider',
  createdAt: '2015-12-17T21:55:54.784Z',
  modifiedAt: '2015-12-17T21:55:54.784Z',
  providerId: 'saml',
  ssoLoginUrl: 'http://login.se',
  ssoLogoutUrl: 'http://logout.se',
  encodedX509SigningCert: '-----BEGIN CERTIFICATE-----...-----END CERTIFICATE-----',
  requestSignatureAlgorithm: 'RSA-SHA1',
  attributeStatementMappingRules: {
    href: 'http://localhost:7001/v1/attributeStatementMappingRules/7MPwgswI85ZzR3RYFJ1KIm'
  },
  serviceProviderMetadata: {
    href: 'http://localhost:7001/v1/samlServiceProviderMetadatas/7NB10jttftaGM1ygGOZOei'
  }
}
```

lib/resource/Directory.js
```diff
@@ -75,7 +77,19 @@ Directory.prototype.getProvider = function getProvider(/* [options,] callback */
...
+    var providerType = 'Provider';
+
+    if (provider.providerId === 'saml') {
+      providerType = 'Saml';
```
timothyej: Shouldn't this be `'SamlProvider'` instead of just `'Saml'`?

typerandom: Yeah, you're correct. This is wrong. Going to fix that :)

lib/resource/Directory.js
typerandom: Can you share the code that you're using in a gist?

timothyej: https://gist.github.com/timothyej/74d9523da136406b9f77

The problem is that it doesn't recognise the `providerFactory` function as a *Resource constructor*. It must be a constructor of `Resource`. See `isAssignableFrom` in `utils.js`.

timothyej: This can be solved like this:

```javascript
function providerFactory(provider) {
  Tenant.super_.apply(this, arguments);
  //...
}
utils.inherits(providerFactory, InstanceResource);
```

timothyej: ...but that doesn't really solve it either. It's not possible to send in a factory with the current structure.

timothyej: I guess you meant to send invoke it as `providerFactory(this.provider)` instead of sending in the function itself?

timothyej: I've fixed this for you. Please review my commit (dcae147b12c3b1886d0676d919363aa2d8e8e539).

lib/saml/SamlIdpUrlBuilder.js
```diff
@@ -0,0 +1,46 @@
...
+
+SamlIdpUrlBuilder.prototype.build = function (callback) {
+  var self = this;
+
+  this._getServiceProvider(function (err, serviceProvider) {
+    if (err) {
+      return callback(err);
+    }
+
+    var accessToken = njwt.create({
```
timothyej: According to the [spec](https://stormpath.atlassian.net/wiki/display/AM/Stormpath+as+a+SAML+Service+Provider), the JWT header should include the API Key ID in the field `kid`.

lib/saml/SamlIdpUrlBuilder.js
```diff
@@ -0,0 +1,46 @@
...
+    if (err) {
+      return callback(err);
+    }
+
+    samlPolicy.getServiceProvider(callback);
+  });
+};
+
+SamlIdpUrlBuilder.prototype._buildInitializationUrl = function _buildInitializationUrl(initEndpointUrl, accessToken) {
+  return initEndpointUrl + '?accessToken=' + encodeURIComponent(accessToken);
```
timothyej: According to the [spec](https://stormpath.atlassian.net/wiki/display/AM/Stormpath+as+a+SAML+Service+Provider), the URL should also be able to include these optional parameters:

`accountStore.href`, `accountStore.nameKey`, `callbackUri`, `state`

lib/saml/SamlIdpUrlBuilder.js
timothyej: I just fixed this for you in my commit (c7f882e3e599f357c60b7905fcfc738a07410ddd).

lib/saml/SamlIdpUrlBuilder.js
timothyej: I can't get this URL to work, I only get `401` responses:

```javascript
{
    status: 401,
    code: 401,
    message: "Authentication required.",
    developerMessage: "Authentication with a valid API Key is required.",
    moreInfo: "http://www.stormpath.com/docs/quickstart/connect"
}
```

lib/authc/StormpathAssertionAuthenticator.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+var njwt = require('njwt');
+var AssertionAuthenticationResult = require('./AssertionAuthenticationResult');
+
+function StormpathAssertionAuthenticator(dataStore) {
```
timothyej: It's not very nice to have to send in the `DataStore` here. It should accept an `Application`  instead, and then you can get the datastore by `application.dataStore`.

timothyej: I fixed this for you in a new commit (bf379c508ea6e4b8d2fed8c91aee38afccc0c457).

lib/authc/AssertionAuthenticationResult.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
```
timothyej: Any special reason why this file is here and not in the `lib/resource/` directory, just like `AuthenticationResult.js`?

lib/authc/AssertionAuthenticationResult.js
```diff
@@ -0,0 +1,37 @@
...
+
+var nJwt = require('njwt');
+var Account = require('../resource/Account');
+
+function AssertionAuthenticationResult(dataStore, data) {
```
timothyej: `AuthenticationResult` inherits `InstanceResource` and also have a few different properties and methods. Is this intended to not be the same or why is this different?

timothyej:
Except for my comments, I've verified that this works. We just need to address these things and then make sure to document it.

lib/saml/SamlIdpUrlBuilder.js
typerandom: Do you want to post the URL that you generated?

lib/saml/SamlIdpUrlBuilder.js
typerandom: Thanks! :)

lib/saml/SamlIdpUrlBuilder.js
timothyej: Sure:

```
https://api.stormpath.com/v1/applications/7NHpJJPzzBbu8ii1kdkD4w/saml/sso/idpRedirect?accessToken=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiIsImtpZCI6IjRDMjNPMVRZMlNCQjdUTUJONEpDT01SWUQifQ.eyJqdGkiOiI1NDEzYmZhMi1jNTE0LTQyMWItYWYwOC1lNGNjNjM0Yzg4OTMiLCJpc3MiOiJodHRwOi8vbG9jYWxob3N0OjcwMDEvdjEvYXBwbGljYXRpb25zLzdOSHBKSlB6ekJidThpaTFrZGtENHciLCJpYXQiOjE0NTI4NzE1MDguNTg1LCJleHAiOjE0NTI4NzUxMDh9.tp5yrgVm7t-T_uR2sVGBORl84GQEY2c1kxntl_W0P74
```

lib/authc/AssertionAuthenticationResult.js
typerandom: It's just a authentication result. I don't understand how that would be connected to a resource. I think it's also strange that AuthenticationResult lies where it do, since that itself isn't a resource.

lib/authc/AssertionAuthenticationResult.js
typerandom: See comment below :)

lib/authc/AssertionAuthenticationResult.js
timothyej: I see, but why can't why just use the `AuthenticationResult` instead of creating a new one?

lib/saml/SamlIdpUrlBuilder.js
timothyej: By bad, something went wrong when I used the Node proxy. When I used the API directly everything worked.

The optional params to the URL still needs to be fixed though :)

timothyej: I just fixed this in my commit 431f79f57ae49879c927f3e16b59cbd4a7333e8a. I'm not totally sure about how the `accountStore` params are entered. Please review and come with suggestions.

lib/authc/AssertionAuthenticationResult.js
typerandom: Good question. I actually based it off of JwtAuthenticationResult, and it felt like AuthenticationResult maybe contained more than was needed. @robertjd What do you think?

typerandom:
Big thanks to @timothyej for being awesome and helping me out with reviewing/fixing this! :beers:

lib/authc/StormpathAssertionAuthenticator.js
```diff
@@ -0,0 +1,42 @@
...
+    // https://docs.stormpath.com/rest/product-guide/latest/005_auth_n.html#step-5-stormpath-response-with-jwt
+    var accountHref = jwt.body.sub;
+
+    if (!accountHref) {
+      return callback(new Error('Stormpath Account HREF (sub) in JWT not provided.'));
+    }
+
+    callback(null, new AssertionAuthenticationResult(
+      dataStore, {
+        jwt: token,
```
robertjd: The `AssertionAuthenticationResult` is expecting this to be called `token`.

lib/authc/AssertionAuthenticationResult.js
```diff
@@ -0,0 +1,37 @@
...
+      if (key in this) {
+        continue;
+      }
+      this[key] = data[key];
+    }
+  }
+
+  var apiKey = this.dataStore.requestExecutor.options.client.apiKey;
+
+  if (this.token) {
```
robertjd: Why a conditional?  There should always be a token.

lib/authc/AssertionAuthenticationResult.js
```diff
@@ -0,0 +1,37 @@
...
+    this.token = nJwt.verify(this.token, apiKey.secret);
+    this.account = {
+      href: this.token.body.sub
+    };
+  }
+}
+
+AssertionAuthenticationResult.prototype.getAccount = function getAccount(callback) {
```
robertjd: Please add a test for this method.

lib/authc/AssertionAuthenticationResult.js
robertjd: The fact that `AuthenticationResult` inherits from `InstanceResource` is weird, and I've wanted to fix that.  I agree that `AuthenticationResult` should be moved to the `lib/authc` folder and not inherit from `InstanceResource`.  Further, `AuthenticationResult` was polluted with methods for client Oauth2 credential authentication, these methods should have been on a new class.  See #357 for more insight into the cleanup that needs to happen.  Of everything I just said, we should work on those issues separately from this PR for SAML.

typerandom:
@robertjd Thanks for the feedback! I have now updated the PR according to it

Annie Teng has been requesting this feature. Do you think that we can merge this in and cut a release?

robertjd:
Thanks @typerandom for the updates!  I spent more time with this PR today, doing end-to-end testing with a tenant that has a SAML IdP setup on a directory.  Through that I found some things that needed to be modified.  I added the following work to this branch:

* `SamlIdpUrlBuilder` needs to pass options in the JWT request, per our documentation

* AssertionResult: renamed token to `stormpath_token`, leave it as plaintext.  Yes this is different from the current ID Site result, but it's more accurate per current conventions.

* JWT responses can be an error, so that needs to be handled

* Hide the datastore from enumeration, because it’s ugly when using console.log() on these objects

I also did some work in the tests:

* Break out the tests of the assertion authentication result, to their own file

* Remove “Called with” assertions on the datastore get resource stub, that’s not necessary IMO

As such, the following still needs to be done on this branch:

* Figure out why the push tests are failing on travis (a rebase may help here, given that we merged #358 today)

* Create documentation pages for the `StormpathAssertionAuthenticator` and `SamlIdpUrlBuilder`, so that it's ready for customer usage.

I have to leave this for the day, but if you could pick this up that would be awesome <3

typerandom:
@robertjd Big thanks for clearing those things up! A rebase fixed the build errors. And I have added some docs. What do you think about those? Do you think that having them as separate sections? Is it the right solution, or should they be put somewhere else?

robertjd:
Thanks!  I added some more context to the docs, merging now!

PR #353, opened 2016-01-16
Move assignments from describe into before/beforeEach

### LINKED ISSUE #334: Resource tests not running properly
When running `npm tests` not all `sp.resource.*` tests run properly.

timothyej:
Fixes #334.

In many of the tests, there are assignments inside the `describe` blocks, which is not very good if they're used in deeper `describe` blocks. For example, in `sp.resource.group_test.js`:

```javascript
describe('Group resource', function () {
  var dataStore = new DataStore({client: {apiKey: {id: 1, secret: 2}}});

  describe('add account', function () {
    describe('if group href not set', function () {
      var group = new Group(dataStore);
      ...
    });
    ...
  });
  ...
});
```

Instead, the assignments should be done in `before`/`beforeEach` blocks while the declarations should be in the `describe` blocks:

```javascript
describe('Group resource', function () {
  var dataStore;

  beforeEach(function () {
    dataStore = new DataStore({client: {apiKey: {id: 1, secret: 2}}});
  });

  describe('add account', function () {
    describe('if group href not set', function () {
      var group;

      beforeEach(function () {
        group = new Group(dataStore);
      });
      ...
    });
    ...
  });
  ...
});
```

You can read more about it here: [Variable in outer describe block is undefined when accessing in inner describe block with Mocha test](http://stackoverflow.com/questions/21470349/variable-in-outer-describe-block-is-undefined-when-accessing-in-inner-describe-b)

@typerandom, can you review and verify, please?

### DIFF
--- a/test/it/account_it.js
+++ b/test/it/account_it.js
@@ -12,10 +12,10 @@ var CustomData = require('../../lib/resource/CustomData');
 var AccountAccessTokenFixture = require('../fixtures/account-token');
 
 describe('Account', function() {
-
-  var fixture = new AccountAccessTokenFixture();
+  var fixture;
 
   before(function(done) {
+    fixture = new AccountAccessTokenFixture();
     fixture.before(done);
   });
 
@@ -90,10 +90,13 @@ describe('Account', function() {
 
       describe('when saved and re-fetched', function() {
         var customDataAfterGet;
-        var propertyName = helpers.uniqId();
-        var propertyValue = helpers.uniqId();
+        var propertyName;
+        var propertyValue;
 
         before(function(done) {
+          propertyName = helpers.uniqId();
+          propertyValue = helpers.uniqId();
+
           customData[propertyName] = propertyValue;
 
           customData.save(function(err) {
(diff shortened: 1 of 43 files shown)


timothyej:
I might have an idea why the coveralls fails. I could be that it is including the actual test code, and when we add more tests, the coverage decreases because of more test code.

typerandom:
Good work! :+1:

It's a lot to look through... But I have checked most of it, it looks consistent and tests are passing. I also fully agree with the reason for the change. Some things can seem meaningless, like:

```javascript
var groupHref;

before(function () {
  groupHref = 'boom2!';
});
```

But for consistency and isolation of tests, I'm good with that.

typerandom:
Also, regarding Coveralls, I'm not really convinced that is the issue. Because the coverage % have changed other times, even when not touching tests. And also, if tests weren't excluded then I'd think that the coverage would be far lower than it is.

timothyej:
Thanks :)

Yes, some of my changes may seem meaningless but like you said, it's better to be consistent and then we will hopefully prevent bugs like this in the future.

Regarding the coverage, I started to look into it and yes, it's probably not the issue.

Issue #150, opened 2015-03-16
Collection Deletion


rdegges:
There's an issue that occurs when iterating over collections, performing delete operations.

Let's say you have an Application with 1,000 accounts, and you want to iterate over each account, deleting it.

You might write code like:

```javascript
application.getAccounts(function(err, accounts) {
  accounts.each(function(account, cb) {
    account.delete();
  }, function(err) {
    console.log('Deleted all accounts!');
  });
});
```

If you were to run this, however, you'd quickly see that this does not, in fact, remove all accounts. Instead it only removes *some* of them due to a bug in our pagination logic. As accounts are deleted, the page numbers are never updated, so we never finish iterating over all accounts.

This should be resolved so any delete() operations successfully impact pagination headers.

robertjd:
This should be investigated as part of #105

Issue #105, opened 2014-10-20
Revisit the implementation of underscore-esque methods


robertjd:
Collection resources have underscore-like methods for iterating, mapping, reducing, etc.  Revisit the implementation of this in order to:
1. Verify that all underscore methods are properly implemented, per the latest version of underscore
2. Verify that our documentation is clearly explaining how to use these methods

typerandom:
@robertjd Is this still relevant or can this issue be closed?

robertjd:
This still needs to be done.  I want someone to interrogate this:

https://github.com/stormpath/stormpath-sdk-node/blob/master/lib/underscore.js

And this:

https://github.com/stormpath/stormpath-sdk-node/blob/master/lib/resource/CollectionResource.js#L155

To make sure that our API for asynchronous iteration of collection is making sense AND is properly documented (we had some bugs in the past, where this was not true)

robertjd:
See the referenced issue, #150, as it may be one of the bugs that I suspect we have in this code

PR #358, opened 2016-01-19
Fixing bad tests

### LINKED ISSUE #334: Resource tests not running properly

timothyej:
The [Move assignments from describe into before/beforeEach](https://github.com/stormpath/stormpath-sdk-node/pull/353) PR made the tests to actually run, but it revealed more errors and tests that failed which are addressed in this PR.

Fixes #334.

### How to verify

1. Checkout this branch
2. Run `npm test`
3. Verify that all tests run (**665** tests)
4. Verify that no tests fails
5. Review code changes based on my notes below

### Changes

* **`test/sp.cache.cache_test.js`**

  * ba569f6bcc3c54b1700428b9277e8e71c135f1b6 Introduced when I refactored assignments into `before`. `opt` is a variable that is being overwritten and will therefore loose its reference. This fix initially assigns it to an object and then it sets the `cache` property inside `before` instead of overwriting it. This will preserve the reference.

* **Use `before` instead of `beforeEach` as it doesn't seem to work**

  * 49cb3b5c6ba391a64896a92ae17a84caf578805d Introduced when I refactored assignments into `before`. There were some problems using `beforeEach` so I changed it to `before` instead. In later refactors I was able to use `beforeEach`, so I might go back and check this fix later. I would prefer to use `beforeEach` rather than `before` for better isolation.

* **Custom errors should capture the stack trace in a proper way**

  * c6c84f28a901fa88ea63315574b2fe80f5861986 I found this when I was looking for another bug, and I decided to fix this as I found it. Basically I improved the way the error stack is captured. You can read more there: <https://gist.github.com/justmoon/15511f92e5216fa2624b>.

* **`Error.prepareStackTrace` should be restored even when the filename is found, otherwise Mocha will crash on an error**

  * a277ba0d591d220621c0b3c28c4cc8ec95d4b67d Whenever a test failed, Mocha crashed complaining that `stack.split` was undefined. I debugged this and eventually found that is was because the `Error` object had been modified without being restored due to a bug.

* **`test/sp.cache.entry_test.js`**

  * b42e837e8e432ebafa14c1a7247d64b7f430d8e5 This test was failing because it was generating a timestamp and then directly after checked if some time has passed. This might have worked on slow computers, but not mine ;) I fixed this by adding a timeout of 20 ms.

* **`test/sp.config.configLoader_test.js`**

  * 9787fe2964a8b5819986395db0a7908cf1ba3caf I replaced all random uuids with hardcoded ones instead. Since one of the uuid were leaking into the other tests, this made it a lot more easier to find, as I could just search for the uuid that didn't match.

  * fca6aff5ca3be88a5eb72404f5a027b7a2ce1af3 Tests weren't isolated and one config was not removed and made other tests fail. I fixed this by using `beforeEach` and `afterEach` to setup and restore the state between each test.

  * fca6aff5ca3be88a5eb72404f5a027b7a2ce1af3 I also had to mock the home path with a `~`, and not just expanded as the `fake-fs` didn't seem to handle that. 

* **`u.BASE_URL` should not contain `/v1` as this will be added by `u.v1()`**

  * caf90aa7f5518cbe96650e11233afa5faed8059a This made a lot of tests fail as it would result in URLs such as `https://api.stormpath.com/v1/v1/applications/etc`.

* **`.calledWith(undefined, undefined)` does not equal *no arguments*, instead I replaced it with `.calledWithExactly()`**

  * cc8c60c0e0e7fb4e57add51048a263296f7b371f As the title says, `.calledWith(undefined, undefined)` does not equal without any arguments. For this assertion to be true, the spy would had to be called with two undefined arguments. I also renamed the test descriptions from `'should call cb without options'` to `'should call cb without arguments'`.

* **`test/sp.resource.collectionResource_test.js`**

  * af640c06fddd5fd59744a1de5fcfb7fa18f36bf9 The iterator should return `true`/`false` based on the method that is under test.

  * 12955bc2d92517053c6cdc49e48ae5aed98d1168 Should only test if the callback is called once, not other things (which is not even correct).

  * f6f9b9a3a8d678061f8108c315be90e8518cd8a2 Test all methods with pagination.

* **`lib/resource/CollectionResource.js`**

  * 8d8f3a939682009a6e26cff0d7565eaa5e088af0 `async.every(arr, iterator, [callback])` will call the callback with `true` as result if all truth tests passed. Many of the other methods will be the opposite (calling it with `false`). This will make `async.parallel()` to believe that an error occurred as it just sends its result directly into the `parallel` callback. This caused `.every()` to behave very odd and it called its callback between each page.

   * 4d9bf5b10489f8ceffee120b24c92b4196c76320  `.every(arr, iterator, [callback])` should abort as soon as an iterator returns `false`.

* **`test/sp.resource.directory_test.js`**

  * 091308ae090b0d615c96d1bb5d36961d83fa0df6 Use `u.v1()` to prepend the version to the URL. Also make sure to pass the error to `done()` so we know when an error occurs.

  * 4997807da56b4af9cd846e24be48bd9aa1e24dd1 `provider` should be checked against `providerObj`, and not itself.

  * 8baae7e30a6458bf4e5b9d592a664f1eb2060680 `organizationMappings` should be `organizations`.

* **`test/it/oauth_id_site_token_grant_it.js`**

  * 455559e8cfacc87fe92a6fe98eb816c6549945f4 Remove empty test which will otherwise get stuck in `pending`.

* **Increase timeouts to prevent Travis from timing out when running tests with a lot of old resources**

  * c668c9441be05a10445f48f67b8bd88731b02049, c1d84301af9a4ec4c8ccce05a2fb24c0600e050d, 3d83ff0f4477bbd9d9bfd7f123728bdf28f9e2f7 It seems like some tests doesn't clean up after themselves, and that leads to a lot of resources which eventually makes the tests to timeout. I won't address this issue in this PR, so I just increased the timeout for those tests.

### DIFF
--- a/test/it/tenant_it.js
+++ b/test/it/tenant_it.js
@@ -9,7 +9,7 @@ var Tenant = require('../../lib/resource/Tenant');
 describe('Tenant', function() {
   var client, tenant;
 
-  this.timeout(20000);
+  this.timeout(120000);
 
   before(function(done) {
     helpers.getClient(function(_client) {
--- a/test/it/tenant_it.js
+++ b/test/it/tenant_it.js
@@ -9,7 +9,7 @@ var Tenant = require('../../lib/resource/Tenant');
 describe('Tenant', function() {
   var client, tenant;
 
-  this.timeout(120000);
+  this.timeout(60 * 2 * 1000);
 
   before(function(done) {
     helpers.getClient(function(_client) {
--- a/test/it/tenant_it.js
+++ b/test/it/tenant_it.js
@@ -9,7 +9,7 @@ var Tenant = require('../../lib/resource/Tenant');
 describe('Tenant', function() {
   var client, tenant;
 
-  this.timeout(10000);
+  this.timeout(20000);
 
   before(function(done) {
     helpers.getClient(function(_client) {
--- a/test/it/tenant_it.js
+++ b/test/it/tenant_it.js
@@ -9,6 +9,8 @@ var Tenant = require('../../lib/resource/Tenant');
 describe('Tenant', function() {
   var client, tenant;
 
+  this.timeout(10000);
+
   before(function(done) {
     helpers.getClient(function(_client) {
       client = _client;
--- a/test/sp.ds.requestExecutor_test.js
+++ b/test/sp.ds.requestExecutor_test.js
@@ -13,7 +13,7 @@ describe('ds:', function () {
   describe('RequestExecutor:', function () {
     var apiKey;
 
-    beforeEach(function () {
+    before(function () {
       apiKey = {id: 1, secret: 2};
     });
 
@@ -31,7 +31,7 @@ describe('ds:', function () {
       describe('create with required options', function () {
         var reqExec;
 
-        beforeEach(function () {
+        before(function () {
           reqExec = new RequestExecutor({client: {apiKey: apiKey}});
         });
 
@@ -57,7 +57,7 @@ describe('ds:', function () {
         };
       }
 
-      beforeEach(function () {
+      before(function () {
         reqExec = new RequestExecutor({client: {apiKey: apiKey} });
       });
 
--- a/test/sp.ds.requestExecutor_test.js
+++ b/test/sp.ds.requestExecutor_test.js
@@ -13,7 +13,7 @@ describe('ds:', function () {
   describe('RequestExecutor:', function () {
     var apiKey;
 
-    this.timeout(10000);
+    this.timeout(10 * 1000);
 
     before(function () {
       apiKey = {id: 1, secret: 2};
--- a/test/sp.ds.requestExecutor_test.js
+++ b/test/sp.ds.requestExecutor_test.js
@@ -13,6 +13,8 @@ describe('ds:', function () {
   describe('RequestExecutor:', function () {
     var apiKey;
 
+    this.timeout(10000);
+
     before(function () {
       apiKey = {id: 1, secret: 2};
     });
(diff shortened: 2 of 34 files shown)


timothyej:
@typerandom @robertjd @rdegges, could you please review? :) And let me know if you want me to break this up into smaller PRs.

test/it/tenant_it.js
```diff
@@ -9,6 +9,8 @@ var Tenant = require('../../lib/resource/Tenant');
 describe('Tenant', function() {
   var client, tenant;
 
+  this.timeout(120000);
```
typerandom: Really minor thing. But for readability I'd rather write this as `60 * 2 * 1000`.

test/sp.ds.requestExecutor_test.js
```diff
@@ -13,7 +13,9 @@ describe('ds:', function () {
   describe('RequestExecutor:', function () {
     var apiKey;
 
-    beforeEach(function () {
+    this.timeout(10000);
```
typerandom: Same as previous comment.

typerandom:
@timothyej Besides a really minor comment. I think this looks good! Good work! :+1:

test/it/tenant_it.js
timothyej: Thanks! (3d83ff0f4477bbd9d9bfd7f123728bdf28f9e2f7)

robertjd:
This looks great, many thanks to @timothyej for digging into these issues!  IMO this also closes #105.  It probably fixes #150 as well, but someone will need to check.

PR #359, opened 2016-01-21
Release 0.15.5

robertjd:
See CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 * Fixed: `collection.every(iterator, callback)`.  The callback was being invoked
   on every page request to the API (it should only be called once the entire
-  collection has been paginated).
+  collection has been paginated).  The iteration would be aborted if the
+  iterator returned `false`.  This is now fixed, and the iterator continues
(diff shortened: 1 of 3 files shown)


rdegges:
+1

timothyej:
:+1:

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # stormpath-sdk-node Change Log
 
+### 0.15.5
+
+**Released January 21, 2016**
+
+* Fixed: `collection.every(iterator, callback)`.  The callback was being invoked
+  on every page request to the API (it should only be called once the entire
+  collection has been paginated).  The iteration would be aborted if the
```
timothyej: > The iteration would be aborted if the
> iterator returned `false`.  This is now fixed, and the iterator continues
> regardless of the return value of the iterator (as `every` should do).

Actually the opposite! Before, it didn't break no matter the return value. Now I've made it break on `false`:

<https://github.com/caolan/async#every>

> `callback(result)` - Optional A callback which is called as soon as any iterator returns `false`, or after all the iterator functions have finished. Result will be either `true` or `false` depending on the values of the async tests.

PR #360, opened 2016-01-22
Release 0.16.0 - SAML Support

robertjd:
See changelog


timothyej:
:+1:

robertjd:
Thanks for the review, as mentioned in chat: two lines have lost their coverage.  I saw this happen once yesterday as well, it seems to be random.  If this continues to happen we should investigate the tests that cover these lines.

Issue #11, opened 2014-03-14
Auth fails


amjibaly:
https://github.com/stormpath/stormpath-sdk-node/blob/master/lib/ds/DataStore.js#L149

When authenticating (using expand=account), body.href is undefined. You need to use body.account.href.

PR #362, opened 2016-01-25
Feature: Add OAuth authenticator for stormpath_token

rorheden:
Adds the `OAuthStormpathTokenGrantAuthenticator` for authenticating `stormpath_token` tokens.

Part solution of: https://github.com/stormpath/express-stormpath/issues/251

### DIFF
--- a/test/it/oauth_authenticator_it.js
+++ b/test/it/oauth_authenticator_it.js
@@ -159,6 +159,24 @@ describe('OAuthAuthenticator',function(){
     });
   });
 
+  it('should error when trying to authenticate with invalid stormpath token', function(done){
+    var authenticator = stormpath.OAuthAuthenticator(application);
+    authenticator.authenticate({
+      body: {
+        grant_type: 'stormpath_token',
+        stormpath_token: 'abc'
+      }
+    },function(err,result){
+      assert.isUndefined(result);
+
+      assert.isNotNull(err);
+      assert.equal(err.status, 400);
+      assert.equal(err.code, 10017); // 10017 = Token is invalid.
+
+      done();
+    });
+  });
+
   it('should error when trying to authenticate with invalid refresh token', function(done){
     var authenticator = stormpath.OAuthAuthenticator(application);
     authenticator.authenticate({
--- a/test/it/oauth_stormpath_token_it.js
+++ b/test/it/oauth_stormpath_token_it.js
@@ -1,0 +1,36 @@
+
+var common = require('../common');
+var helpers = require('./helpers');
+var assert = common.assert;
+
+var stormpath = require('../../');
+
+describe('OAuthStormpathTokenAuthenticator', function () {
+  var newAccount;
+  var application;
+
+  before(function(done){
+    newAccount = helpers.fakeAccount();
+
+    helpers.createApplication(function (err, app) {
+      application = app;
+      application.createAccount(newAccount, done);
+    });
+  });
+
+  after(function (done) {
+    application.delete(done);
+  });
+
+  it('should be constructable with new operator',function(){
+    var authenticator = new stormpath.OAuthStormpathTokenAuthenticator(application);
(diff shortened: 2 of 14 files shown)


test/it/oauth_authenticator_it.js
```diff
@@ -154,6 +154,24 @@ describe('OAuthAuthenticator',function(){
...
+      body: {
+        grant_type: 'stormpath_token',
+        stormpath_token: 'abc'
+      }
+    },function(err,result){
+      assert.isUndefined(result);
+
+      assert.isNotNull(err);
+      assert.equal(err.status, 400);
+      assert.equal(err.code, 10017); // 10017 = Token is invalid.
```
rdegges: Should we potentially define these constants somewhere? eg

```javascript
var TOKEN_IS_INVALID = 10017;
```

rdegges:
Left a comment, but looks good to me.

Also: since we decided to use ES6, this might be a good place to do it. EG: You have the perfect opportunity to define a new style class =)

robertjd:
Hey @typerandom , thanks for this!  We need to do some more work here:

* We can't remove the existing `OAuthIdSiteTokenGrantAuthenticator`, as that's a breaking change.  That needs to be brought back, but should receive a deprecation notice via `util.deprecate()`

* It turns out that we didn't finish our naming work for these new authenticators, so I'm pushing that conversation forward on [sdk-spec #11](https://github.com/stormpath/stormpath-sdk-spec/pull/11).  Please take a look at my most recent comment to get  sense of where we are going with this.  We need to resolve that issue and update this PR accordingly.

robertjd:
Alright!  We resolved our spec around the authenticators today, please review the latest here so that we can determine how this PR needs to change:

https://github.com/stormpath/stormpath-sdk-spec/blob/master/specifications/authenticators.md

test/it/oauth_authenticator_it.js
typerandom: Yeah, that would be a good idea. I think I'll look into doing a separate refactor for that.

typerandom:
@robertjd This is also updated to comply with the spec.

robertjd:
Thanks, the implementation looks correct now.  Please fix the tests to have 100% on the new `OAuthStormpathTokenAuthenticator`, and add a documentation page for this authenticator.  We reference this in the deprecation notice so it must be documented as the new way of validating ID Site token callbacks.  Back to @typerandom

test/it/oauth_stormpath_token_it.js
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  it('should be constructable with new operator',function(){
+    var authenticator = new stormpath.OAuthStormpathTokenAuthenticator(application);
+    assert.instanceOf(authenticator, stormpath.OAuthStormpathTokenAuthenticator);
+    assert.equal(authenticator.application, application);
+  });
+
+  it('should be constructable without new operator',function(){
+    var authenticator = new stormpath.OAuthStormpathTokenAuthenticator(application);
```
timothyej: I think you meant to delete `new` here?

> should be constructable **without new** operator

typerandom: Thanks! Fixed that.

typerandom:
@robertjd @timothyej Thanks for your comments. I've updated the PR with some more tests and docs. Please review again.

robertjd:
Tests look great now, thanks!  I pushed two commits to fix small issues that I saw:

* Documentation needed to mention that this new authenticator can be used for SAML as well.
* Documentation needed some more padding in the sidebar.
* The way that `OAuthIdSiteTokenGrantAuthenticator` and `OAuthIdSiteTokenGrantAuthenticationResult ` were deprecated, by wrapping the constructor, is technically a breaking change because you're no longer returning the same type.  As such I moved the deprecation notice to the `authenticate()` method, which will suffice for fair warning (you wouldn't use either class without calling that method) and won't break the API.

Merging!

PR #378, opened 2016-01-28
Add `Organization.getDefaultAccountStore()` and `Organization.getDefaultGroupStore()` methods

timothyej:
This PR implements the comment from PR #340:

> * `Organization.getDefaultAccountStore()` and `Organization.getDefaultGroupStore()` will return the account store that is identified by the mapping.  When doing the first step of fetching the mapping, use the option to expand the account store.  This will get that resource cached locally, so that when you make the next request to get the account store, it will already be cached.
> * `Organization.getDefaultAccountStoreMapping()` and `Organization.getDefaultGroupStoreMapping()` need to be implemented and will return the account store mapping.  
> – @robertjd 

### How to verify

1. Checkout this branch
2. Create a new file called `org.js` in the root
3. Paste this content into the file:
  ```javascript
'use strict';

var stormpath = require('./lib');

var apiKey = new stormpath.ApiKey(
  '<KEY ID>',
  '<KEY SECRET>'
);

var client = new stormpath.Client({ apiKey: apiKey });
var organizationHref = 'https://api.stormpath.com/v1/organizations/<id>';

client.getOrganization(organizationHref, function(err, organization) {
  organization.getDefaultAccountStore(function (err, accountStore) {
    console.log('account store:', err, accountStore);
  });

  organization.getDefaultGroupStore(function (err, groupStore) {
    console.log('group store:', err, groupStore);
  });
});
```
4. Insert your own key id and secret
5. Also insert an *Organization* href
6. Run `node org.js`
7. Verify that the output is correct (should be the *Directories* that are the default account and group stores)

### DIFF
--- a/lib/resource/Organization.js
+++ b/lib/resource/Organization.js
@@ -7,6 +7,7 @@ var Account = require('./Account');
 var CustomData = require('./CustomData');
 var IdSiteModel = require('./IdSiteModel');
 var OrganizationAccountStoreMapping = require('./OrganizationAccountStoreMapping');
+var Directory = require('./Directory');
 
 function Organization() {
   Organization.super_.apply(this, arguments);
@@ -47,6 +48,32 @@ Organization.prototype.getDefaultGroupStoreMapping = function getDefaultGroupSto
   }
 
   return this.dataStore.getResource(this.defaultGroupStoreMapping.href, args.options, OrganizationAccountStoreMapping, args.callback);
+};
+
+Organization.prototype.getDefaultAccountStore = function getDefaultAccountStore(/* [options,] callback */) {
+  var args = utils.resolveArgs(arguments, ['options', 'callback'], true);
+  var dataStore = this.dataStore;
+
+  return this.getDefaultAccountStoreMapping({ expand: 'accountStore' }, function (err, accountStoreMapping) {
+    if (err) {
+      return args.callback(err);
+    }
+
+    dataStore.getResource(accountStoreMapping.accountStore.href, args.options, Directory, args.callback);
+  });
+};
+
+Organization.prototype.getDefaultGroupStore = function getDefaultGroupStore(/* [options,] callback */) {
+  var args = utils.resolveArgs(arguments, ['options', 'callback'], true);
+  var dataStore = this.dataStore;
+
+  return this.getDefaultGroupStoreMapping({ expand: 'accountStore' }, function (err, groupStoreMapping) {
+    if (err) {
+      return args.callback(err);
+    }
+
+    dataStore.getResource(groupStoreMapping.accountStore.href, args.options, Directory, args.callback);
+  });
 };
 
 Organization.prototype.getGroups = function getGroups(/* [options,] callback */) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/resource/Organization.js [hunk 1]
```diff
@@ -49,6 +50,32 @@ Organization.prototype.getDefaultGroupStore = function getDefaultGroupStore(/* [
   return this.dataStore.getResource(this.defaultGroupStoreMapping.href, args.options, OrganizationAccountStoreMapping, args.callback);
 };
 
+Organization.prototype.getDefaultAccountStore = function getDefaultAccountStore(/* [options,] callback */) {
+  var args = utils.resolveArgs(arguments, ['options', 'callback'], true);
+  var dataStore = this.dataStore;
+
+  return this.getDefaultAccountStoreMapping({ expand: 'accountStore' }, function (err, accountStoreMapping) {
```
robertjd: `accountStoreMapping` should be `organizationAccountStoreMapping`, as it is that type.

lib/resource/Organization.js
```diff
@@ -49,6 +50,32 @@ Organization.prototype.getDefaultGroupStore = function getDefaultGroupStore(/* [
...
+    if (err) {
+      return args.callback(err);
+    }
+
+    dataStore.getResource(accountStoreMapping.accountStore.href, args.options, Directory, args.callback);
```
robertjd: The account store can be a group or directory, so this cannot be hard-coded as a directory.

`OrganizationAccountStoreMapping` inherits from `AccountStoreMapping`, which has a `getAccountStore()` method that can resolve the appropriate account store type.  I'd suggest replacing this `data.getResource` call with `organizationAccountStoreMapping.getAccountStore()`

timothyej:
@robertjd great suggestions :) I've addressed them in my last two commits. Please review again :)

lib/resource/Organization.js
```diff
@@ -49,6 +49,30 @@ Organization.prototype.getDefaultGroupStore = function getDefaultGroupStore(/* [
...
 
+Organization.prototype.getDefaultAccountStore = function getDefaultAccountStore(/* [options,] callback */) {
+  var args = utils.resolveArgs(arguments, ['options', 'callback'], true);
+
+  return this.getDefaultAccountStoreMapping({ expand: 'accountStore' }, function (err, organizationAccountStoreMapping) {
+    if (err) {
+      return args.callback(err);
+    }
+
+    organizationAccountStoreMapping.getAccountStore(args.options, args.callback);
```
robertjd: There needs to be a null check here: `organizationAccountStoreMapping` will be `null` if there is no default account store mapping for the organization.  In this situation we need to callback with (null,null).  Right now it throws an exception because `getAccountStore` is undefined on null.  Back to @timothyej

timothyej:
@robertjd This has now been fixed (ab721af09c857f5890067cbefe0c5779a8dd8fed). Please review (@typerandom too :)).

typerandom:
Looks good! :+1:

robertjd:
Thanks @timothyej , this is ready to merge now.  Let's continue a conversation about how we want tests to be written in this library.

PR #383, opened 2016-02-05
Fix: Handle thrown JWT errors

typerandom:
Fixes an issue where `njwt.verify()` throws an error that isn't being handled, and therefore crashes the server.

##### How to verify
1. Checkout this branch.
2. Run `$ npm link` to make the `stormpath` module available for linking.
3. Checkout master in `express-stormpath`.
4. Run `$ npm link` to make the `express-stormpath` module available for linking.
5. Run `$ npm link stormpath` to link to the `stormpath` module with our patch.
6. Create an new Node.js server using [this gist](https://gist.github.com/typerandom/8e97e235aaeac1fae840) (in a new directory).
7. Run `$ npm link express-stormpath` to link to the `express-stormpath` module.
8. Start the server.
9. Once `Server listening on http://localhost:3000/` is shown, navigate to [http://localhost:3000/](http://localhost:3000/).
10. Refresh the page multiple times and assert that the messag "Invalid refresh token cookie set" is being shown.
11. Assert that the server is still running and that there are no console errors.

Fixes https://github.com/stormpath/express-stormpath/issues/275

### DIFF
--- a/lib/jwt/jwt-authenticator.js
+++ b/lib/jwt/jwt-authenticator.js
@@ -36,39 +36,43 @@ JwtAuthenticator.prototype.authenticate = function authenticate(token,cb){
 
   var secret = self.application.dataStore.requestExecutor.options.client.apiKey.secret;
 
-  njwt.verify(token,secret,function(err,jwt){
-    if(err){
-      err.statusCode = 401;
-      cb(err);
-    }else{
-      if(self.localValidation){
-        cb(null, new JwtAuthenticationResult(self.application,{
-          jwt: token,
-          expandedJwt: jwt,
-          localValidation: true,
-          account: {
-            href: jwt.body.sub
-          }
-        }));
-      }else if(jwt.header.kid){
-        // If the KID exists, this was issued by our API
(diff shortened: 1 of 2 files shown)


timothyej:
Using the verification steps, I get the same result on both `master` and this branch. No console errors or anything.

timothyej:
Another note: Shouldn't you add the same error handling to all places that are using the `njwt.verify()` method? E.g. `StormpathAssertionAuthenticator` and `JwtAuthenticationResult`.

timothyej:
A third note: Maybe add a test that tests that it handles invalid JWTs? :)

typerandom:
@timothyej I'm actually going to look into this a bit more. Since there seems to maybe be something more at play... I.e. we're actually getting a token from the API... But for some reason it's not returning as valid (it seems).

robertjd:
LGTM, thanks @typerandom !

robertjd:
P.S. yes, the cause for getting an invalid token error should be investigated, but I +1 this fix because it's wrong to throw errors when you are passed a callback.  The callback should be used instead.

PR #393, opened 2016-02-22
Set user agent header on every request

typerandom:
Fixes so that the user agent is resolved/set on each request. This enables us to change the config and provide a different user agent in runtime.

#### How to verify

1. Checkout this branch.
2. Create a new test application using [this gist](https://gist.github.com/typerandom/de31c20ac1701ffc5a9e) that you link to this branch.
2. Set the `userAgent` config of `new stormpath.Client()` to something of e.g. `foo/1.2.3`.
3. Add a `console.log(options)` underneath line `64` in `lib/ds/RequestExecutor.js`.
4. Run the test application.
5. Assert that the `options.headers` object contains a `User-Agent` with the value that you've set.

### DIFF
--- a/lib/ds/RequestExecutor.js
+++ b/lib/ds/RequestExecutor.js
@@ -26,9 +26,6 @@ function RequestExecutor(options) {
   this.requestAuthenticator = authc.getAuthenticator(options);
 
   options.headers = options.headers || {};
-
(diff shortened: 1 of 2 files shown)


robertjd:
+1, looks good!

PR #415, opened 2016-03-18
First pass at JsDoc

robertjd:
Hey @timothyej and @typerandom !

So here it is, my first pass at re-writing the documentation with JsDoc.  It's not complete, but I wanted us to review the structure now.  I focused on the Application and Account resources, and while doing that I branched out to some other resources.  Please look at how I'm using JsDoc and let me know your thoughts!

To build the docs:

```
npm run docs
```

The output will go into the `apidocs/` folder, then you can open the files in your browser.

Or if you have `nodemon` and `httpster` installed as global modules, you can use this cool script that will start a fileserver and rebuild the docs as you make changes:

```
./docs-dev.sh
```

EDIT: I've also added my 1.0 changes document, this is where I'm keeping a list of things that we need to clean up in the API, and these things should happen with our 1.0 release because they will be breaking changes.  We can break these out to issues when we start that project, but I wanted to at least keep track of them for now.


typerandom:
This is awesome. I'm going to review it more in-depth tomorrow :)

lib/resource/SamlPolicy.js
```diff
@@ -3,15 +3,56 @@
...
+ * @param  {Function} callback
+ * Callback function, will be called with (err, {@link SamlServiceProvider samlServiceProvider}).
+ */
 SamlPolicy.prototype.getServiceProvider = function getServiceProvider(/* [options,] callback */) {
   var args = utils.resolveArgs(arguments, ['options', 'callback'], true);
   return this.dataStore.getResource(this.serviceProvider.href, args.options, require('./SamlServiceProvider'), args.callback);
 };
 
+/**
+ *
```
timothyej: tumbleweed

robertjd: :)

typerandom:
@robertjd Hey! I have reviewed this and I think it looks good. Made a few changes, but mostly just styling and fixes. Not anything that rewrote the docs.

@timothyej Could you also review this?

timothyej:
I've now been reviewing all of it, awesome work @robertjd! :+1:

This is what I've been fixing in my commit (3673bb9beb0a71d19c3fd600cb824434e48d0013):

* Typos
* Missing punctuation
* Corrected some of the docs (parameters, missing class names, etc.)

I also fixed some other things based on my preference:

* In comments, write markdown lists with `-` instead of `*` as the comment itself uses `*`.
* When using dummy domains, use `example.com` rather than `mydomain.com` etc. as these domains might exist.
* Use only one whitespace after period. Since one space seems to be the majority and Robin removed some double whitespace, I also went with this.
* Wrap lines longer than ~80 characters.
* Moved class descriptions to the top.

I also noticed that some class documented its constructor, but while most don't.

robertjd:
Thanks everyone!  The edits are great and I'll use this format going forward.

robertjd:
I'm going to close this PR, as I was using it for initial review.  Work is continuing on this branch, I will make another squashed PR when done!

PR #434, opened 2016-03-30
Don't cache created accounts

### LINKED ISSUE #433: Unverified accounts are being cached
Creating an account with email verification turned on will leave the account unverified if the user verifies his account through another server. E.g. via the default link `
https://api.stormpath.com/emailVerificationTokens?sptoken=<token>`.

### How to reproduce

1. Turn on Email Verification on your Application
2. Run this gist: https://gist.github.com/timothyej/29271aaf85fbb8c88346b76f7a230087
3. Verify that the returned account after authentication have its status set to `UNVERIFIED`

or

1. Turn on Email Verification on your Application
2. Run the `dashboard-app` example project from the AngularJS SDK and signup for a new account
3. Click the link in the email (`
https://api.stormpath.com/emailVerificationTokens?sptoken=<token>`)
4. Try to login
5. Verify that the server still thinks that the account is unverified (even though the admin console says the opposite)
6. Restart the Express server and verify that you now can log in

### Diagnostics

* The account is cached from when it was registered
* The flow works fine on the `express-stormpath-sample-project`

### Screenshots

screen shot 2016-03-30 at 13 29 32

timothyej:
This PR solves the problem with accounts still being unverified even though they actually were verified. The proposed solution is to not cache created accounts, and thus getting a fresh resource at login.

Fixes #433.

### Discussion

Is this the right way to go? We could be more specific and cache all accounts unless they're unverified.

### How to verify

1. Turn on Email Verification on your Application
2. Checkout `master`
3. Save this Gist as `test.js` in the root: https://gist.github.com/timothyej/29271aaf85fbb8c88346b76f7a230087
4. Replace `var stormpath = require('stormpath');` with `var stormpath = require('./lib');`
5. Run `node test.js`
6. Verify that the returned account has its status set to `UNVERIFIED`
7. Checkout this branch and run `node test.js` again
8. Verify that the account this time has its status set to `ENABLED`

### DIFF
--- a/lib/resource/Application.js
+++ b/lib/resource/Application.js
@@ -231,7 +231,23 @@ Application.prototype.getAccount = function getAccount(/* providerData, [options
 
 Application.prototype.createAccount = function createApplicationAccount(/* account, [options,] callback */) {
   var args = utils.resolveArgs(arguments, ['account', 'options', 'callback']);
-  return this.dataStore.createResource(this.accounts.href, args.options, args.account, require('./Account'), args.callback);
+  var dataStore = this.dataStore;
+
+  return dataStore.createResource(this.accounts.href, args.options, args.account, require('./Account'), function (err, createdAccount) {
+    var callbackArguments = arguments;
+
+    if (err) {
+      return args.callback(err);
+    }
+
+    dataStore._evict(createdAccount.href, function (err) {
+      if (err) {
+        return args.callback(err);
+      }
+
+      args.callback.apply(null, callbackArguments);
+    });
+  });
 };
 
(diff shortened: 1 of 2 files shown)


robertjd:
I'm not a huge fan of this change, I regularly inform people that resources go into the cache when they are created.  I don't think this should be a special case.  If we want to address #433 we should do it through the framework integration, documentation, or even a configuration warning.

timothyej:
Thanks for the input, @robertjd. I will think about this one a bit. One thing could also be to make a note about it in the Admin Console for that email template.

PR #449, opened 2016-04-18
Fix .save() named function in Group.js

MrMaz:
Prototype function `group.save()` points to an incorrectly named `saveAccount` function.

This is trivial and only affects stack traces.


robertjd:
Thanks @MrMaz !  We actually removed this child implementation of save() in #457 so this patch is no longer necessary.

PR #457, opened 2016-05-16
Fix organization.save() for custom data

### LINKED ISSUE #455: Organization.save() error when customData populated.
Hello!

I've run into an issue with trying to save an organization that has customData populated via expansion.

```javascript
client.getOrganization(href, {expand: 'customData'}, function(err, organization){
  organization.name = 'Foo';
  organization.save();
});
```

This gives me the following error:

> Property name 'createdAt' is invalid.  Property names may not equal any of the following reserved names: 'href', 'createdAt', 'modifiedAt', meta', 'spMeta', 'spmeta', 'ionmeta', or 'ionMeta'.

Making a similar request for an account or group does not seem to be an issue because they extend the `DirectoryChildResource`. Organization directly extends `InstanceResource` which doesn't do any of the cleanup of reserved fields before saving.

I was on 0.17.5 when I first encountered this, and upgrading to 0.18.2 did not resolve it.

My current workaround is to copy the customData property, delete it from the object, then re-assign it to the object after save.

Please let me know if you need any further info.

Thanks :)

typerandom:
Fixes the issue where protected custom data fields weren't removed before saving the Organization.

#### How to verify

1. Create a new app using [this gist](https://gist.github.com/typerandom/376a13f3e7ddc40aeac9d3c1c110f5a9).
2. Test the app against the Node SDK's master branch.
3. Verify that you get the error described in #455.
2. Test the app against this branch (`fix-organization-custom-data-save`).
5. Verify that every thing works and the custom data for the specified organization is updated.

Fixes #455

### DIFF
--- a/lib/resource/Account.js
+++ b/lib/resource/Account.js
@@ -95,9 +95,7 @@ Account.prototype.save = function saveAccount(){
     });
   }
 
-  self._applyCustomDataUpdatesIfNecessary(function(){
-    Account.super_.prototype.save.apply(self, args);
-  });
+  Account.super_.prototype.save.apply(self, args);
 };
 
 module.exports = Account;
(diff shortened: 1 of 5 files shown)


robertjd:
Looks good!

PR #483, opened 2016-06-21
Added cache invalidation to verifyAccountEmail.

blyork:
Repost of stormpath/stormpath-sdk-node#481 in order to change target branch.
Fix for issue stormpath/express-stormpath#448.

PR #484, opened 2016-06-21
Added cache invalidation to verifyAccountEmail.

robertjd:
Re-opened with new test, new api keys.  Closes #483

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,5 @@ sudo: false
 script: travis_retry npm test
 env:
   global:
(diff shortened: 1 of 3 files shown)


robertjd:
This PR will be merged with test failures, the test failures are unrelated to the changes in this PR.  The test failures are happening on the api key resource endpoints, we have a bug in the REST API that needs to be fixed (it's not an issue with this library).  These api key tests will pass again once the bug is resolved.

PR #485, opened 2016-06-21
Release 0.18.3

robertjd:
See CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # stormpath-sdk-node Change Log
 
+### 0.18.3
+
+**Released June 22, 2016**
+
+* `Tenant.verifyAccountEmail()` will now evict the account from the cache before
+  the operation.  Previously it would replace the cache entry after the
+  operation, which can be problematic (stormpath/express-stormpath#448).
+
+* Fixed bug where `Organization.save()` would error if custom data had been
+  expanded (#455).
+
+* Updated production dependencies:
+
+ * `memcached@2.2.1` -> `memcached@2.2.2`
+ * `moment@2.12.0` -> `moment@2.13.0`
+ * `request@2.69.0` -> `request@2.72.0`
+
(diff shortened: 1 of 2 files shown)


robertjd:
Note for code review: Test errors regarding "api key name" errors should be ignored, there is a bug in the REST API that is causing the problem and it is currently being worked on.  Please inspect the tests to make sure that the only errors are relating to api keys.  If this is true, we are safe to merge with failing tests.

timothyej:
This is the result from the tests:

```console
784 passing (3m)
  4 failing

  1) Application.authenticateApiRequest with a disabled api key "before all" hook:
     Uncaught HTTP 400, Stormpath 2002 (http://docs.stormpath.com/errors/2002): apiKey name is an invalid value.


  2) Account api keys disable an api key "before all" hook:
     Uncaught TypeError: Cannot read property 'href' of null
      at test/it/api_keys_it.js:168:59
      at onSaveResourceRequestResult (lib/ds/DataStore.js:9:19723)
      at Request.onRequestResult [as _callback] (lib/ds/RequestExecutor.js:9:4385)
      at Request.self.callback (node_modules/request/request.js:200:22)
      at Request.<anonymous> (node_modules/request/request.js:1067:10)
      at IncomingMessage.<anonymous> (node_modules/request/request.js:988:12)
      at endReadableNT (_stream_readable.js:913:12)
      at _combinedTickCallback (internal/process/next_tick.js:74:11)
      at process._tickCallback (internal/process/next_tick.js:98:9)

  3) OAuthAuthenticator should reject expired tokens:
     Uncaught Jwt is expired


  4) OAuthStormpathTokenAuthenticator when calling authenticate(data, callback) with a stormpath_token data property and the token is valid should succeed with a OAuthStormpathTokenAuthenticationResult result:
     Uncaught AssertionError: expected { Object (name, status, ...) } to be falsy
```

Other than that it looks fine :)

PR #500, opened 2016-07-13
Fix issue when comparing config loaders of different node module versions

typerandom:
Fixes the issue where the module versions being used in the Express SDK and Stormpath Node SDK are out of sync and `stormpath-config` module versions differ.

##### How to verify

Will be added soon.

Fixes https://github.com/stormpath/express-stormpath/issues/467
Fixes https://github.com/stormpath/express-stormpath/issues/463

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -2,7 +2,6 @@
 
 var async = require('async');
 var events = require('events');
-var stormpathConfig = require('stormpath-config');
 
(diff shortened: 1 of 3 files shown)


robertjd:
I've confirmed that this fixes the reported problem.  I've rebased this branch against master, where I fixed some error message strings in the test files.  Tests are good now.

PR #503, opened 2016-07-20
Release 0.18.4

robertjd:
See CHANGELOG.med

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # stormpath-sdk-node Change Log
 
+**Released July 20, 2016**
+
+### 0.18.4
+
+* Fixed the client constructor to properly accept configuration loader instances
+  that are passed into this library.  Fixes stormpath/express-stormpath#463.
+
 ### 0.18.3
 
 **Released June 22, 2016**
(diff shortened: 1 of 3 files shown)


timothyej:
I've added some missing entries (`0.18.1` ... `0.18.4`) to UPGRADING.md and I also added a notice about the updated dependencies to CHANGELOG.md.

But there seem to be some problem with previous releases. GitHub flags `0.17.5` as the latest:

screen shot 2016-07-20 at 13 02 10

And when expanding the *3 newer tags* it is still missing `0.18.3`:

screen shot 2016-07-20 at 13 03 11

Anything I'm missing here?

robertjd:
Thanks @timothyej ! I went ahead and enumerated the updated packages, like I usually do :)

Regarding Github releases, you have to do those manually (it's a Github feature that isn't tied to git repo data).  The 0.18.3 tag wasn't pushed from my machine, that's my bad.

I've been thinking that we need to start automating some of our release processes. I recently started working on a script that will fail the tests if the PR is a release PR and doesn't have the necessary elements in the changelog, upgradelog, package.json, etc.  I'll send that out for review when I have more time to work on it :)

PR #507, opened 2016-07-26
Fix .save() named function in Group.js

robertjd:
Prototype function `group.save()` points to an incorrectly named `saveAccount` function.

This is trivial and only affects stack traces.

Re-opened on this repo so that tests can run, thanks @MrMaz!  Closes #449


robertjd:
Oh, my bad.  We actually removed the child implementation of Group.save() in #457, so this patch is no longer necessary.  Closing this and #449.

PR #508, opened 2016-07-26
Add requestId and request context to resource errors

robertjd:
Adding more information to our error objects.  The requestId is a new value that comes down on REST API responses.  The requested URL and method are generally useful for debugging.  I specifically chose URL (not URI) because we deal with canonical Stormpath resource URLs in this request executor.

### DIFF
--- a/lib/ds/RequestExecutor.js
+++ b/lib/ds/RequestExecutor.js
@@ -84,7 +84,7 @@ RequestExecutor.prototype.execute = function executeRequest(req, callback) {
     }
 
     if (response.statusCode > 399) {
-      return callback(new ResourceError(body || {status:response.statusCode}), null);
+      return callback(new ResourceError(body || {status:response.statusCode}, {url: req.uri, method: req.method}), null);
     }
 
     if (response.statusCode === 201){
--- a/lib/ds/RequestExecutor.js
+++ b/lib/ds/RequestExecutor.js
@@ -77,6 +77,7 @@ RequestExecutor.prototype.execute = function executeRequest(req, callback) {
   this.requestAuthenticator.authenticate(options);
 
   request(options, function onRequestResult(err, response, body) {
+    var responseContext = this;
     if (err) {
       var wrapper = new Error('Unable to execute http request ' + req.method + ' ' + req.uri + ': ' + err.message);
       wrapper.inner = err;
@@ -84,7 +85,7 @@ RequestExecutor.prototype.execute = function executeRequest(req, callback) {
     }
 
     if (response.statusCode > 399) {
-      return callback(new ResourceError(body || {status:response.statusCode}, {url: req.uri, method: req.method}), null);
+      return callback(new ResourceError(body || {status:response.statusCode}, {url: responseContext.href, method: req.method}), null);
     }
 
     if (response.statusCode === 201){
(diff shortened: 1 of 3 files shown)


lib/ds/RequestExecutor.js
```diff
@@ -84,7 +84,7 @@ RequestExecutor.prototype.execute = function executeRequest(req, callback) {
     }
 
     if (response.statusCode > 399) {
-      return callback(new ResourceError(body || {status:response.statusCode}), null);
+      return callback(new ResourceError(body || {status:response.statusCode}, {url: req.uri, method: req.method}), null);
```
typerandom: Are you really meaning `url` here and not `uri`? Because the value that I received was `/tenants/current`. Which is not a URL.

lib/error/ResourceError.js
```diff
@@ -2,15 +2,20 @@
...
+  requestData = requestData || {};
+
   this.name = this.constructor.name;
   this.status = responseBody.status;
   this.code = responseBody.code;
   this.userMessage = responseBody.message;
   this.developerMessage = responseBody.developerMessage;
   this.moreInfo = responseBody.moreInfo;
+  this.requestId = responseBody.requestId;
+  this.url = requestData.url;
```
typerandom: See other comment. Should this really be `url` and not `uri`?

typerandom:
I have tested this and it works as expected. I'm also happy with the code. But I left a comment, so just wondering if the value there is right, or if we should change the name from `url` to `uri`.

lib/ds/RequestExecutor.js
robertjd: I do want to have the full URL logged, can you show me how you produced the `/tenants/current` value? When I tested this, I simply made a request for an application that doesn't exist:

```

var applicationHref = 'https://api.stormpath.com/v1/applications/foo';

client.getResource(applicationHref,function (err) {
  console.log(err);
});
```

And I get this error, which gives me the full URL:

```
{ [ResourceError: HTTP 404, Stormpath 404 (mailto:support@stormpath.com): The requested resource does not exist.]
  name: 'ResourceError',
  status: 404,
  code: 404,
  userMessage: 'The requested resource does not exist.',
  developerMessage: 'The requested resource does not exist.',
  moreInfo: 'mailto:support@stormpath.com',
  requestId: '8d80cd20-534b-11e6-a56f-22000bc0820c',
  url: 'https://api.stormpath.com/v1/applications/foo',
  method: 'GET',
  message: 'HTTP 404, Stormpath 404 (mailto:support@stormpath.com): The requested resource does not exist.' }
```

typerandom: This is the code I'm using:

```javascript

  client: {
    apiKey: {
      id: 'invalid id',
      secret: 'invalid secret'
    },
  }
});

client.on('stormpath.error', function (err) {
  console.log('Stormpath error:', err);
});

client.getCurrentTenant(function (err, tenant) {
  console.log(err, tenant);
});
```

robertjd:
@typerandom I've updated this to always report the full URL, using the `herf` property that the request library is providing on the response context.  I've also added a test.  Please take one last look before I merge?  Thanks for helping me find this case.

typerandom:
@robertjd Fixed some spacing. Otherwise looks good! 👍 So once the build show all greens I'm going to merge it in.

PR #514, opened 2016-08-03
keepalive

devpascoe:
Because we are performing a lot of handshaking and multiple SSL urls in my testing adding this option to keepalive the connection shaved 5 seconds off initial startup time.

### DIFF
--- a/lib/ds/RequestExecutor.js
+++ b/lib/ds/RequestExecutor.js
@@ -74,6 +74,8 @@ RequestExecutor.prototype.execute = function executeRequest(req, callback) {
     options.json = true; // All Stormpath resources are JSON
   }
 
+  options.forever = true;
+
   this.requestAuthenticator.authenticate(options);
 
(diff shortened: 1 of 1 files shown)


robertjd:
Thanks Steve,  I’ve followed up through our support channel, but for posterity I’ll put the same info here.

The keep alive option will reduce some time on that initial SSL handshake, so if you’re making a lot (thousands) of request per second, this may make sense.  Otherwise I advise against it, because we do have a load balancer in front of our service, and there is no certainty about when it will terminate your connection, thus introducing other failure states.

As such, I do not want to merge this as the default.  But if you’d still like to use the option, you can pass it down through our Client constructor like this:

```
new stormpath.Client({
  forever: true
});
```

This is a nuance about how configuration is parsed in this current version.  We’re putting together our roadmap for our 1.0 release, which will include a better interface for configuring (or providing) the request executor.

Thanks again for looking into this!

PR #531, opened 2016-08-19
Adding Browserify support

### LINKED ISSUE #529: Look into how we can browserify the Stormpath SDK

edjiang:
Fixes #529. 

Seems to actually work well, from my testing. The only thing *not* working, from what I can tell so far is the `/tenants/current` endpoint, but it looks like it's because the proxy isn't forwarding a needed header (since I can reproduce in Paw)

### DIFF
--- a/lib/configLoader.js
+++ b/lib/configLoader.js
@@ -9,6 +9,10 @@ var stormpathPath = '~/.stormpath';
 
 // Create a default client config loader.
 module.exports = function (extendWithConfig) {
+  // If in a browser, do the minimal to load config. Since this is only used in IAM-UI, 
+  // no need to validate. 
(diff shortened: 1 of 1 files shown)


edjiang:
I figured it out -- the reason it's not working is that the proxy is trying to follow the redirect, doesn't send the authorization header, then gets a redirect to /login which it proxies back to the client.

edjiang:
@robertjd I realized the tenants/current bug was because I wasn't putting /v1 in the baseURL

edjiang:
This is being closed in favor of the injectable strategy that Robert's taking in https://github.com/stormpath/stormpath-node-config/tree/request-executor-fix

PR #534, opened 2016-08-28
Add Directory.getAccountSchema()

### LINKED ISSUE #526: Account Schema Configuration
https://stormpath.atlassian.net/wiki/display/AM/Loosening+Data+Model+Requirements AM-2067

typerandom:
Adds support for retrieving the account schema for a directory.

#### How to verify

Test the new SDK API with the following code:

```javascript

client.getDirectory('DIRECTORY HREF HERE', function (err, directory) {
  if (err) {
    return console.error(err);
  }

  directory.getAccountSchema(function (err, accountSchema) {
    if (err) {
      return console.error(err);
    }

    accountSchema.getFields(function (err, fields) {
      if (err) {
        return console.error(err);
      }

      fields.each(function (field, next) {
        console.log('Found field!', field);
        next();
      });
    });
  });
});
```

Fixes #526

### DIFF
--- a/lib/resource/Directory.js
+++ b/lib/resource/Directory.js
@@ -77,6 +77,21 @@ Directory.prototype.getAccountCreationPolicy = function getAccountCreationPolicy
 };
 
 /**
+ * Get the {@link Schema} resource of this Directory resource.
+ *
+ * @param {ExpansionOptions} [expansionOptions]
+ * For retrieving linked resources of the {@link Schema} during this request.
+ *
+ * @param {Function} callback
+ * Callback function, will be called with (err, {@link Schema}).
+ */
+Directory.prototype.getAccountSchema = function getAccountSchema(/* [options,] callback */) {
+  var args = utils.resolveArgs(arguments, ['options', 'callback'], true);
+  return this.dataStore.getResource(this.accountSchema.href, args.options, require('./Schema'), args.callback);
+};
+
+
+/**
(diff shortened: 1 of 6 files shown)


robertjd:
Hey @typerandom I’ve added some important changes to this branch.  Please take a review of these:

* The `Schema` resource now inherits from `Resource`, because this resource cannot be saved or deleted.  As such we should not expose those methods, which can be achieved by inheriting from `Resource` instead of `InstanceResource`.

* I created a specific `Field` resource for the same reason: this resource can be saved, but not deleted.  I’ve duplicated a `save()` method on the `Field` prototype to achieve this.

* I’ve added more documentation, linked to the full documentation, and provided an example that shows the developer how to change a field requirement (as this is what the Schema is for).  All new PRs should receive this amount of documentation love :)

Regarding the second point, I spent some time looking for a mixin solution that would let use re-use the `save()` method where needed.  I couldn’t find a mixin solution that I liked for pre-ES6 code.  But I did find this solution that I really like for ES6 code, and I think we should try implementing this when we convert this library to ES6: http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/

typerandom:
@robertjd Thanks. I'm happy about the changes so going to merge this in. We'll have to continue the discussion on how to do mixins:) Not quite sure which one I think is the best option.

PR #535, opened 2016-09-08
Add Client href error handling

### LINKED ISSUE #516: Better Error Messages
I'm going through our console quickstart for node. The quickstart shows using environment variables to set important application credentials.

If I don't set these variables, and attempt to run the code samples, I get an error that looks like this:

```javascript
/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/ds/RequestExecutor.js:53
    throw new Error('request.uri field is required.');
    ^

Error: request.uri field is required.
    at RequestExecutor.executeRequest [as execute] (/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/ds/RequestExecutor.js:53:11)
    at doRequest (/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/ds/DataStore.js:272:27)
    at onCacheResult (/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/ds/DataStore.js:296:5)
    at Array.<anonymous> (/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/cache/Cache.js:54:14)
    at DisabledCache.get.DisabledCache.set.DisabledCache.delete.DisabledCache.clear.DisabledCache.size (/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/cache/DisabledCache.js:11:62)
    at Cache.get (/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/cache/Cache.js:52:14)
    at CacheHandler.getCachedResource [as get] (/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/cache/CacheHandler.js:91:51)
    at Object.executeRequest [as exec] (/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/ds/DataStore.js:289:22)
    at DataStore.getResource (/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/ds/DataStore.js:117:16)
    at Client.getResource (/Users/rdegges/Dropbox/Code/testing/samples/node_modules/stormpath/lib/Client.js:189:38)
```

This looks pretty nasty. Would be nice to error out better with a usable message.

typerandom:
Improves the Client prototype by validating methods requiring a href to be a valid href.

#### How to verify

Using the code below and the `master` branch, verify that you receive the verbose `request.uri field is required` error as seen in issue #516. Then using the `add-client-error-handling` branch, use the same code and verify that you receive the error `Argument 'href' (undefined) is not a valid Application href`. If `STORMPATH_APPLICATION_HREF` is set in your environment, then simply set the variable `applicationHref` to `undefined` to replicate the error.

```javascript

var apiKey = new stormpath.ApiKey(
  process.env['STORMPATH_CLIENT_APIKEY_ID'],
  process.env['STORMPATH_CLIENT_APIKEY_SECRET']
);

  apiKey: apiKey
});

var applicationHref = process.env['STORMPATH_APPLICATION_HREF'];

client.getApplication(applicationHref, function(err, application) {
  if (err) {
    return console.error(err);
  }

  console.log('Got application!', application);
});
```

**Note:** The code above is the same that you end up with when you go through the Node.js quickstart in the [Stormpath console](https://api.stormpath.com/ui2/index.html#/quickstart/none/nodejs/backend/project-type/existing).

#### Discussion

As you can see in the code I've added some sanity checking on the hrefs being provided. E.g. a call to `getApplication()` must now provide a href which is a valid URI and has a path containing `/applications/`. The question I'm wondering is, what do you think of this, and are there cases (resources) where we don't want to be so strict about the URI being passed?

Fixes #516

### DIFF
--- a/lib/error/invalid-href.js
+++ b/lib/error/invalid-href.js
@@ -1,0 +1,16 @@
+'use strict';
+
+/**
+ * Produces an Erorr object with a message that indicates which resource type has
+ * violated an href validation.
+ *
+ * @param      {string}  href          The href that was passed into a resource getter.
+ * @param      {string}  resourceName  The human readable name of the expected resource type.
+ * @return     {Error}   A simple error with the message
+ */
+function InvalidHrefError(href, resourceName){
+  var message = 'Argument \'href\' (' + href + ') is not a valid ' + resourceName +' href.';
+  return new Error(message);
+}
+
+module.exports = InvalidHrefError;
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -5,6 +5,7 @@ var events = require('events');
 
 var utils = require('./utils');
 var DataStore = require('./ds/DataStore');
+var InvalidHrefError = require('./error/invalid-href');
 var ObjectCallProxy = require('./proxy/ObjectCallProxy');
 
 /**
@@ -304,7 +305,7 @@ Client.prototype.getAccessToken = function () {
   var args = utils.resolveArgs(arguments, ['href', 'options', 'callback']);
 
   if (!utils.isValidHref(args.href, '/accessTokens/')) {
-    return args.callback(new Error('Argument \'href\' (' + args.href + ') is not a valid Access Token href.'));
+    return args.callback(new InvalidHrefError(args.href, 'Access Token'));
   }
 
   return this.getResource(args.href, args.options, require('./resource/AccessToken'), args.callback);
@@ -334,7 +335,7 @@ Client.prototype.getRefreshToken = function () {
   var args = utils.resolveArgs(arguments, ['href', 'options', 'callback']);
 
   if (!utils.isValidHref(args.href, '/refreshTokens/')) {
-    return args.callback(new Error('Argument \'href\' (' + args.href + ') is not a valid Refresh Token href.'));
+    return args.callback(new InvalidHrefError(args.href, 'Refresh Token'));
   }
 
   return this.getResource(args.href, args.options, require('./resource/RefreshToken'), args.callback);
@@ -791,7 +792,7 @@ Client.prototype.getAccount = function () {
   var args = utils.resolveArgs(arguments, ['href', 'options', 'callback']);
 
   if (!utils.isValidHref(args.href, '/accounts/')) {
-    return args.callback(new Error('Argument \'href\' (' + args.href + ') is not a valid Account href.'));
+    return args.callback(new InvalidHrefError(args.href, 'Account'));
   }
 
   return this.getResource(args.href, args.options, require('./resource/Account'), args.callback);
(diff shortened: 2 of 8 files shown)


typerandom:
@robertjd 

I fully agree that doing it DRY is important and that was actually my first approach. The issue I had with just doing error handing in the `RequestExecutor` is that the `Client` and `RequestExecutor` are really two different domains.

The `Client` methods take a `href` argument and the `RequestExecutor.execute()` method takes a `req.uri` argument. Design-wise, it would be confusing to refer to `href` in the request executor and you'd also lose a bit of information in doing so. So the reason why I did inline error handling/guarding is because it added more value and context to each call. I also think it's good practice to guard against errors as early as possible. Even if it added LOC and wasn't the most DRY approach, in my opinion it made it much easier to trace what call that actually threw the error.

I.e. compare the error "Argument 'href' (null) is invalid." compared to "Argument 'href' (null) is not a valid Application href.".

The first error could be thrown from literally any method calling the request executor if the href would be invalid. But the second error you'd be certain that it was related to your call to e.g. `Client.getApplication()`. So in that sense it would be a lot easier for the user to understand such an error message and debug it. Of course, there are other solutions. And a little more DRY approach would probably be to have the error messages wrapped in `Client.getResource()`.

So what do you think about changing the `Client.getResource()`-method so that instead of just forwarding the call to `_dataStore.getResource()` it would have the following signature instead `getResource(resourceName, requiredSubPath, href, options, callback)?

I.e. the previous `getResource(args.href, args.options, require('./resource/Application'), args.callback)` would instead look like `getResource('Application', '/applications/', args.href, args.options, args.callback)`.

robertjd:
Thanks @typerandom - you're right, my proposed changes would defeat the purpose of this PR, which is to make it more obvious that the app href isn't being passed correctly.  As a compromise I've created an `InvalidHrefError` function to reduce the number of repeated error message strings :) merging now.

lib/error/invalid-href.js
```diff
@@ -0,0 +1,16 @@
...
+ * Produces an Erorr object with a message that indicates which resource type has
+ * violated an href validation.
+ *
+ * @param      {string}  href          The href that was passed into a resource getter.
+ * @param      {string}  resourceName  The human readable name of the expected resource type.
+ * @return     {Error}   A simple error with the message
+ */
+function InvalidHrefError(href, resourceName){
+  var message = 'Argument \'href\' (' + href + ') is not a valid ' + resourceName +' href.';
+  return new Error(message);
```
typerandom: @robertjd I really like the approach of having an `InvalidHrefError`! But a question. Any reason why you return a new `Error` from a constructor rather than have the `InvalidHrefError` extend from Error? I find returning things from a constructor to be bad practice and a little confusing. Mostly since the type you'd expect isn't what you'd get, i.e. `err instanceof InvalidHrefError` would fail. And you allocate two objects instead of just one.

typerandom:
@robertjd I know this is merged. I really like your approach with having a `InvalidHrefError`, but I left a small comment for you. So please check that out and let me know :)

PR #544, opened 2016-09-17
Create OAuthClientCredentialsAuthenticator

### LINKED ISSUE #487: client_credentials grant flow needs to use the REST API
### Goals for this task:

* Create a new `OAuthClientCredentialsAuthenticator` and `OAuthClientCredentialsAuthenticationResult`, which should go in `lib/oauth/client-credentials.js`.  This new authenticator and authentication result will allow developers to use Stormpath to generate access + refresh token pairs for their end-users, after the end-user has authenticated with an API Key pair.

* Create integration tests in a new file `test/it/oauth_client_credentials.js` (the existing `client_credential_auth_it.js` test will be deprecated in the future).  The tests should assert the following:

    * That an access + refresh token can be created for an account, when using the `OAuthClientCredentialsAuthenticator`.
    * That `oAuthClientCredentialsAuthenticationResult.getAccount()` will return the correct account resource.
    * That `oAuthClientCredentialsAuthenticationResult.getAccessToken()` will return the correct access token resource.

### Task description

The Stormpath REST API now supports the client-credential authentication flow.  In this situation, the end-user, who has an Account object, also has an API Key attached to their account.  The end user want to use this API key to authenticate with a service, and the service needs to use the Node SDK to accept the API key and return an access token + refresh token in response.

Please see this documentation to understand the feature, as it exists in the REST API:

https://docs.stormpath.com/rest/product-guide/latest/auth_n.html#using-stormpath-for-token-based-authentication

We have already implemented the password grant flow in the Node SDK, and you can read about it here:

https://docs.stormpath.com/nodejs/jsdoc/OAuthPasswordGrantRequestAuthenticator.html

The Node SDK does support `client_credentials` right now, but it's a confusing API that manages tokens locally (it does not use the Stormpath REST API) and we want to deprecate it, and create a proper `OAuthClientCredentialsAuthenticator` and `OAuthClientCredentialsAuthenticationResult` instead.  The deprecated interface is `Application.authenticateApiRequest()`.

Here is an example of how this new authenticator should be used

```javascript
// Simple usage: just exchange API Keys for some tokens

var application; // retrieved by client.getApplication(href)

var authenticator = new OAuthClientCredentialsAuthenticator(application);

var apiKeyId, secret;  // From an API Key resource that is attached to an Account

var clientCredentialsRequest = {
  apiKey: {
    id: 'WuQvJFCj+MNHy/MldQp/dVu9WjCc7EYl0yjR7fUp2ym',
    secret: '7HB3CD9YN5YXKFWRQEUIPMGZ1'
  }
}

authenticator.authenticate(clientCredentialsRequest, function(err, oAuthClientCredentialsAuthenticationResult) {
    console.log(oAuthClientCredentialsAuthenticationResult.access_token);
});
```

In that example, the `OAuthClientCredentialsAuthenticator` has made a POST to the application's `/oauth/token` endpoint, and sent the end-user's API Keys in the post body.  If the API Keys are valid, the REST API responds with a response that contains an `access_token` and `refresh_token`.

The `OAuthClientCredentialsAuthenticationResult` needs to provide these three methods:

* `getAccount()` - which should fetch the Account object that is related to this response, this can be done by using the `nJwt` library to unpack the `access_token`, and use the `sub` property inside the token (it is the HREF of the account).

* `getAccessToken()` - the response will have a `stormpath_access_token_href`, which can be used to fetch the `AccessToken` resource.

Tweety-FER:
Implements an authenticator that uses the `client_credentials` grant request, and a result encapsulator. Contains tests and documentation for the new authenticator and authentication result. Follows specification outlined in #487 

## Usage Example
```javascript
var application; // retrieved by client.getApplication(href)

var authenticator = new OAuthClientCredentialsAuthenticator(application);

var apiKeyId, secret;  // From an API Key resource that is attached to an Account

var clientCredentialsRequest = {
  apiKey: {
    id: apiKeyId,
    secret: secret
  }
};

authenticator.authenticate(clientCredentialsRequest, function(err, oAuthClientCredentialsAuthenticationResult) {
  oAuthClientCredentialsAuthenticationResult.getAccount(function(err, account) {
    if (err) {
      throw err;
    }
      
    console.log('Account', account);
  });

  oAuthClientCredentialsAuthenticationResult.getAccessToken(function(err, accessToken) {
    if(err) {
      throw err;
    }

    console.log('Access Token', accessToken);
  });
});
```

Fixes #487


robertjd:
FYI all I'm still reviewing this, I want to move the `refreshToken` property from the `JwtAuthenticationResult` to the `PasswordGrantAuthenticationResult`, but doing so causes the express-stormpath tests to fail - I will look into this tomorrow.

robertjd:
Thanks @Tweety-FER !  I made a few small changes to the docs, and renamed one variable name for clarify.  Looks good otherwise.

PR #545, opened 2016-09-19
Rename IdSiteModel to IdSite

### LINKED ISSUE #537: IdSiteModel should be called IdSite
We currently have a resource model that is named `IdSiteModel`. For consistency this should be called just `IdSite`.

Tweety-FER:
Renamed the file and class. Foundnd all references in documents and tests and updated them.

Fixes #537

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -936,7 +936,7 @@ Client.prototype.getOrganizations = function getOrganizations(/* [options,] call
 };
 
 /**
- * Retrieves all the {@link IdSiteModel} resources for the current {@link Tenant}.
+ * Retrieves all the {@link IdSite} resources for the current {@link Tenant}.
  *
  * @param {CollectionQueryOptions} [collectionQueryOptions]
  * Options for querying, paginating, and expanding the collection.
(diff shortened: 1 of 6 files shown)


typerandom:
@Tweety-FER Thanks! Unfortunately after more investigation it looks as if the idSiteModel is "correctly named". I.e. the resource is actually named idSiteModel. Which in my opinion is a little bit confusing, but it seems that the resource name is like that.

@robertjd Do you have any input here?

robertjd:
Ah!  I see where the confusion is here.  There are two resources in the REST API:

* The **ID Site Model** is a view of other data that is fetched from `application/:id/idSiteModel` - it's not a modifiable resource.

* The **ID Site** is a real resource, that contains the configuration information for an ID Site.  This is a real resource that can be updated (but not deleted, there one per tenant right now).

Thus, this PR has some correct changes, but some incorrect changes.  The following should be true:

* `Application.getIdSiteModel()` should return `IdSiteModel`
* `Organization.getIdSiteModel()` should return `IdSiteModel`
* `Tenant.getIdSites()` should return `IdSite`

Ping @Tweety-FER for changes, thanks!

PR #551, opened 2016-09-26
Feature instantiate expanded resources

### LINKED ISSUE #103: Expanded resources are not constructed
Currently: if you expand a resource then that data will be attached to the parent resource as an object literal, not as a constructed instance of the resource type.  As such they do not have the methods you expect.  For example:

```javascript
client.getGroup(href,{expand:directory},function(err,group){
    group.directory.getProvider()  // undefined
});
```

This issue represents the work to ensure references are returned as proper Resource instances rather than object literals.

Tweety-FER:
Implements resource expansion upon object creation through a `ResourceFactory`. When a resource is queried or created, and a valid `{expand: 'some,resource,names'}` key-value pair is attached to the query, these resources will be replaced with an appropriated Resource class instance containing the same data. When call yields a `CollectionResource` instead of a single resource, each instance in the collection will undergo this process.

Usage example:
```
var group; // Retrieved via getGroups() or created earlier

group.getAccountMemberships({expand: 'account'}, function(err, memberships) {
  if (err) {
    throw err;
  }
  
  memberships.each(function(membership, next) {
     var account = membership.account;
     
     console.log(account instanceof Account); // logs `true`

     account.getAccessTokens(function(err, accessTokens) {
       console.log('User', account.username, 'has', accessTokens.items.length, 'access tokens');
       next();
     });
  }, function() {
    console.log('Accounts memberships processed');
  });
});
```

Fixes #103 

I do need your opinion on something, however: I've only written the unit tests for this feature (and did some manual testing to ensure that it works). Do you think I should be adding integration tests for all resource instances that (can) use the `expand` property, i.e. their methods that do? This would certainly be beneficial, but would have to be an extra chuck of code that is maintained and updated every time the API updates (as the current code will continue working even if another `expand` option is added one day).

### DIFF
--- a/lib/resource/ResourceFactory.js
+++ b/lib/resource/ResourceFactory.js
@@ -1,9 +1,35 @@
 'use strict';
-
+/* jshint -W003 */
 var CollectionResource = require('./CollectionResource');
 var CustomData = require('./CustomData');
 var InstanceResource = require('./InstanceResource');
 var utils = require('../utils');
+
+
+function expandResource(expandedFields, resource, query, dataStore) {
+  if (resource instanceof CollectionResource) {
+    resource.items.forEach(function(instance) {
+      expandResource(expandedFields, instance, query, dataStore);
+    });
+
+    return resource;
+  }
+
+  expandedFields.forEach(function(fieldName) {
+    var normalizedFieldName = fieldName.charAt(0).toUpperCase() + fieldName.substr(1);
+    var expandedPath = './' + normalizedFieldName;
+
+    if (typeof resource[fieldName] !== 'undefined') {
+      try {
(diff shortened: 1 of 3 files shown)


robertjd:
Thanks @Tweety-FER ! @typerandom can you please review?  Don't merge yet, I'd like us to have  chance to both look.

typerandom:
@robertjd Yes, I will review this too. Will get back to you when I've done it fully.

typerandom:
I have reviewed this. Made a little code cleanup. But other than that I'm happy about the changes.

Tweety-FER:
@typerandom Thanks and sorry about that. Looks like Atom is drawing upon the wrong config for its linting, and it managed to pass the grunt linter. Will be more careful with that in future PRs.

robertjd:
Looks great, thanks @Tweety-FER !

PR #553, opened 2016-10-08
Remove sub requirement in StormpathAssertionAuthenticator

### LINKED ISSUE #547: StormpathAssertionAuthenticator throws on LOGOUT callbacks
The `StormpathAssertionAuthenticator` throws this error if the callback JTW has no subject field:

'Stormpath Account HREF (sub) in JWT not provided.'

This causes a problem in situations where the user is being logged out via ID Site, and the api.stormpath.com/sso endpoint does not have a cookie which retains the user context.  This can happen when the cookie expires or when the cookie has never been set (happens with SAML login flows).

As such, we should remove this assertion, and the `account` property can be null on the callback response.   The purpose of this authenticator is to verify the signature of the token, and callback with an error if the claims body of the token has an `err` value, so this `sub` check is not necessary.

typerandom:
Fixes so that if a `sub` isn't provided in the JWT being authenticated with the `StormpathAssertionAuthenticator`, it simply returns `null`.

#### How to verify

Use the `StormpathAssertionAuthenticator` and pass in a JWT that does not have a `sub`. Verify that the authentication result returned has the account set to `null`.

Fixes #547

### DIFF
--- a/lib/authc/StormpathAssertionAuthenticator.js
+++ b/lib/authc/StormpathAssertionAuthenticator.js
@@ -44,7 +44,6 @@ function StormpathAssertionAuthenticator(application) {
  * @param {String} tokenRequest.stormpath_token
  * The Stormpath Token, from the ID Site or SAML callback.  This is a compacted JWT string.
  *
- *
  * @param {Function} callback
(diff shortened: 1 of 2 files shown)


robertjd:
Verified that this solves the problem by reproducing the case in #547

Issue #539, opened 2016-09-14
Find resources that incorrectly implement save() or delete()


robertjd:
Most of our concrete resources (Directory, etc) inherit from InstanceResource, which provides save() and delete() methods, but these operations are not supported on every REST resource.  We need to find any concrete resources that have methods which can't actually be used, because the REST API doesn't support them.

PR #554, opened 2016-10-10
Feature mfa

### LINKED ISSUE #548: Multifactor Authentication
The following needs to be implemented to support MFA, woot! :)

`Account.createFactor(newFactor, cb)` - See [Adding an SMS Factor](https://docs.stormpath.com/rest/product-guide/latest/auth_n.html#adding-an-sms-factor).  It should callback with a new `SmsFactor` or `GoogleAuthenticatorFactor`, and these should inherit from a common `Factor` type. Note: do not implement anything special for the `?createChallenge=true` parameter.  Simply adding `newFactor.challenge.message` as a field in the request body will trigger the same behavior.

`Account.getFactors(collectionQueryOptions, cb)` should return a collection of `SmsFactor` or `GoogleAuthenticatorFactor` types

`Factor.getAcccount(getResourceOptions, cb)` should return an `Account`.

`Factor.createChallenge(challenge, cb)` - See [Challenging an SMS Factor](https://docs.stormpath.com/rest/product-guide/latest/auth_n.html#challenging-an-sms-factor).  It should return a new `Challenge` type. 

`Factor.getChallenges(collectionQueryOptions, cb)` should return a collection of `Factor` types

`Factor.getAcccount(getResourceOptions, cb)` should return an `Account`.

`SmsFactor.getPhone(getResourceOptions, cb)` should return a new `Phone` resource.

`Phone.getAcccount(getResourceOptions, cb)` should return an `Account`.

`Challenge.getFactor()` and `Challenge.getAcccount()` should be implemented.

Simply unit tests should be added to keep coverage up, no need for IT tests here.

Tweety-FER:
Implements the resources and logic required for multi-factor authentication. These include `Phone`, `Challenge`, `Factor`, and its concrete versions, `SmsFactor` and `GoogleAuthenticatorFactor`. Implements all the resource methods specified by the task (e.g. `Account#createFactor`). In addition to those, implements `Account#getPhones`, to link to the new `Phone` resource, and `Factor#getMostRecentChallenge`, both of which are specified in the API docs. Covers all of these with tests.

In addition to those, it also specifies:
- `FactorInstantiator` -- Helper class + function that enable polymorphic construction of `SmsFactor` and `GoogleAuthenticatorFactor` resources without requiring changes in the underlying resource fetching/creating code
- `SaveableMixin` -- The extraction of the `ResourceInstance` save method into a mixin. This is done because `Phone` is actually a basic `Resource` that can also be `save()`d (something typically done via inheriting `InstanceResource`). This way functionality is preserved, while code duplication is prevented
- `utils.applyMixin(Ctor, Mixin)` -- A utility function for applying mixin functionality to a class prototype

Usage example:

```
var account;
var factorData;
var challengeData = {
  phone: {
    number: '+12312313123'
  }
};

account.createFactor(factorData, function(err, ftr) {
  // Or just use ftr if there is no error
  account.getFactors(factorData, function(err, factors) {
    if (err) {
      throw err;
    }

    var factor = factors.items[0];

    factor.createChallenge(challengeData, function(err, challenge) {
      if (err) {
        throw err;
      }

      var code = readline.question('What is the code? ');

      challenge.code = code;

      challenge.save(function(err, done) {
        factor.delete(function(err) {
          if (err) {
            throw err;
          }

          console.log('Finished with with status: ', done.status);
        });
      });
    });
  });
});
```

Still **TODO**:
1. ~~Test `utils.applyMixin`~~
2. ~~Check for linter rule breaches~~ _jshint is reporting no errors, and warnings only in code I didn't write. I guess that means it's good._
3. ~~Double-check docs for `options` (which, where, where not) for fetching/creating resources from this PR.~~ _Did minor updates, seems good otherwise_

Fixes #548


Tweety-FER:
Undid applying a mixin to Challenge. Inheriting directly for `Resource` explodes the `DataStore`. Looks like something that will have to be handled through #539

lib/resource/FactorInstantiator.js
```diff
@@ -0,0 +1,69 @@
...
+  var Ctor = getFactorConstructor.apply(this, arguments);
+  var argsWithContext = Array.prototype.slice.call(arguments);
+
+  // Adds an initial parameter. It will be the function context (this)
+  // for the bind call. It does not matter because `new` overwrites, it,
+  // however, so we're just setting it to null here for syntactic reasons.
+  argsWithContext.unshift(null);
+
+  return new (Ctor.bind.apply(Ctor, argsWithContext))();
+}
```
Tweety-FER: Briefly, we need *an* argument in front. We're applying `bind`, and it takes `(thisContext, arg1, arg2, ...)`. We need to add an argument simply to have the correct number of arguments passed (otherwise arg1 would become the context, arg2 would become arg1, etc.). However, what *value* we pass is completely irrelevant, because `new` creates a new context for us. So I'm passing `null` to make it more obvious that the value is unimportant.

robertjd: Ah, I see now, thanks!

robertjd:
FYI all I'm still in progress on this review.  I've added a few commits to fix a few things I've already noticed, I will continue tomorrow.

robertjd:
@Tweety-FER I spent more time with this today.  I added some work, but I need to pass this back to you for some more work:

* I forgot to add it to the original issue, but we need `challenge.verifyCode()`.  I've added this, but it needs test coverage.  Can you please work on this?
* I created separate `createFactor()` methods for the SMS and Google types.  Our documentation doesn't reflect this yet, but we are changing the way that we expect people to challenge google factors.  We now want them to create a challenge resource, and then pose the google authenticator code to that challenge resource.  Can you please add test coverage for this as well?
* Several getter methods were using the wrong approach for parsing arguments.  For getter methods that have no default parameters, we need to parse from the right.  Otherwise it will mistakenly think that the callback is the constructor function.  I discovered this while trying to use these methods in example code.
* A lot of the getter methods are missing the appropriate documentation for the expansion options, can you please add these?  Any resource that has expandable linked resources should have this documentation.

Thanks for all the work so far!  

Because we aren't writing IT tests for most of this feature, I'm doing a lot of manual testing of these new APIs with example code.

Tweety-FER:
@robertjd Updated the tests and docs where missing. Removed options for getting `Phone` resource for `SmsFactor`, where none are actually available. Also noticed that the fields are missing for new resources [here](https://docs.stormpath.com/rest/product-guide/latest/reference.html#searchable-attributes).

The failing test is an integration test I didn't touch (Tenant-related). It seems to be environment related, and not caused by any of my changes. Didn't happen in the PR run, and does not happen for me locally.

robertjd:
Thanks @Tweety-FER for all the work on this one!  I pushed a few more updates, to fill in a few more getter methods, and add some new code we need to support ID Site.  I'm going to merge this now.  There were some random test failures that went away after a re-run, I tried to repro locally but couldn't.  If we see those again we'll have to dig in, they dont appear to be related to any code we touched here.

PR #555, opened 2016-10-12
Add SMTP server resource support

### LINKED ISSUE #528: Create SMTP server resource
See: https://docs.stormpath.com/rest/product-guide/latest/reference.html#smtp-server

As such, I should be able to do the following:

```
client.createSmtpServer(newSmtpServer, function(err, smtpServer){ /* .. */})
tenant.createSmtpServer(newSmtpServer, function(err, smtpServer){ /* .. */})
client.getSmtpServers(function(err, smtpServerCollection){ /* .. */})
tenant.getSmtpServers(function(err, smtpServerCollection){ /* .. */})
```

typerandom:
Add support for the retrieving all and creating new `/smtpServers` resources.

#### How to verify

Test the new methods using the sample application below. The SMTP credentials below should be valid but it's only for a SMTP server sandbox, so they aren't sensitive in any way.

```javascript

var newSmtpServer = {
  name: 'My New SMTP Server!',
  host: 'mailtrap.io',
  username: '35082bfcb8b319',
  password: 'f2d06c51a4b88f',
  port: 25
};

client.createSmtpServer(newSmtpServer, function (err, smtpServer) {
  // Only break on the error if isn't already created (code 13003).
  if (err && err.code !== 13003) {
    return console.error('Error!', err);
  }

  client.getSmtpServers(function (err, smtpServers) {
    if (err) {
        return console.error('Error!', err);
    }

    smtpServers.each(function (server, next) {
      console.log('Found SMTP server!', server);
      next();
    });
  });
});
```

Fixes #528

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -758,6 +758,44 @@ Client.prototype.createOrganization = function createOrganization() {
 };
 
 /**
+ * Creates a new SMTP server resource in the current {@link Tenant}.
+ *
+ * @param {Object} server
+ * The {@link SmtpServer} resource to create.
+ *
+ * @param {ExpansionOptions} [expansionOptions]
+ * Options to expand linked resources on the returned {@link SmtpServer}.
+ *
+ * @param {Function} callback
+ * Callback function, will be called with (err, {@link SmtpServer}).
+ *
+ * @example
+ * var newSmtpServer = {
+ *   name: 'My SMTP Server',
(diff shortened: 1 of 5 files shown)


robertjd:
Thanks @typerandom - it wasn't listed in the original issues but we should also have `client.getSmtpServer()`, so I added that to this.

PR #558, opened 2016-10-20
OAuth authenticator scope factory

### LINKED ISSUE #556: Implement "scope factory" for OAuth authenticators
### Feature description

This feature is meant to support developers who want to use Stormpath for token management (we create the tokens for them, and retain a database of issued tokens), but also want to add custom scope to the access tokens that we issue from the `applications/:appId/oauth/token` endpoint.  Developers want to put custom scope in the access token, because it allows them to quickly do access control checks in their application when validating the access token at a later time.

In this situation, we will "add" the custom scope by creating a new access token, this token will be identical to the original token, except we will add the custom scope as a new claim in the token.  So where a default access token response from our API looks like this:

``` javascript
{
  jti: '3xq96rF12hrlb61YSVQwFd',
  iat: 1476749958,
  iss: 'https://api.stormpath.com/v1/applications/24k7HnDOz4tQ9ARsBtPUN6',
  sub: 'https://api.stormpath.com/v1/accounts/5u8BYZtu09s3yd1XtDYRSo',
  exp: 1476750558,
  rti: '3xq96nuw87YmyGvhGFZfdZ'
}
```

The new access token would look like this:

``` javascript
{
  jti: "uUy2ofmqxZHlG2iEvch53",
  iat: 1476749789,
  iss: "https://api.stormpath.com/v1/applications/24k7HnDOz4tQ9ARsBtPUN6",
  sub: "https://api.stormpath.com/v1/accounts/5u8BYZtu09s3yd1XtDYRSo",
  exp: 1476753388,
  rti: "uUy2lLhwNGJ8Qwr2flQSz",
  scope: "foo"
}
```

In order to pass validation, the new token must be signed with the same signing key as the original token.  This signing key is the Tenant API Key Secret that was used when the original token was created.

To allow the developer to add custom scope, we must add a "scope factory" option to our OAuth authenticators.  The scope factory is a function that is provided by the developer, and we call this function after we have received an access token from the server.  If the developer returns a truthy string value from the scope factory, we must create a new access token, and add the `scope` claim to the body of the new token.  We then replace the original token with the new token on the authentication result. 
### Requirements

The following authenticator needs to support the scope factory:
- OAuthPasswordGrantRequestAuthenticator
- OAuthClientCredentialsAuthenticator
- OAuthStormpathTokenAuthenticator

Instances of these authenticators must provide a `setScopeFactory` function for defining the scope factory function.  They must also provide a `setScopeFactorySigningKey` to pass in the signing key.  The must also accept a new, optional, `scope` parameter.  This is used to indicate what scope the client or end user is requesting to have.

The signature of the scope factory function should be `(authenticationResult, requestedScope, callback)`:
- The `authenticationResult` is the authentication result that you currently recieve when you call `authenticator.authenticateAccount(function(err, authenticationResult))`.  By passing the entire result to the developer, they can fetch the account and decide what to do.
- The `requestedScope` is the scope the the end-user has requested while authenticating.  It is up to the developer to look at the requested scope and decide what scope the user is permitted to have.

This feature should have integration tests (using api.stormpath.com) and we should test the following:
- If a scope factory is provided, assert that the value from the scope factory is added to the access token in the authentication result.
- Assert that all the other fields in the token header and claims body are exactly the same.
- Assert that the new token will pass validation with the `JwtAuthenticator`
- Assert that if a token has scope and passes validation, that that scope will be available as `authenticationResult.expandedJwt.claims.scope`.
- Assert that this also works when `jwtAuthenticator.withLocationValidation()` has been used.
- If the developer provides a scope factory, but not a signing key, we should callback with the error, an error should be provided to the callback of `authenticator.authenticate()`.  We should not throw this error.

See the examples below for more context.
### Example Scope Factory: "Groups as scope"

A common use-case is to apply the account's groups as the scope of the token, as a list of space seperated names.  To achieve that, the developer could provide a function like this:

``` javascript
function groupScopeFactory(authenticationResult, requestedScope, callback){
  authenticationResult.getAccount({expand: 'groups'}, function(err, account){
    if (err) {
      return callback(err);
    }

    var groupNames = account.groups.items.map(function (group) {
      return group.name;
    });

    var grantedScope = groupNames.join(' ');

    callback(null, grantedScope);
  });
}
```
### Example use-case: Express application

Here is a pseudo Express application that demonstrates how this new scope factory can be used to issue scope, and how that scope can be checked on future requests.

``` javascript
var express = require('express');

var tenantApiKey = {
  id: 'stormpath tenant api key id',
  secret: 'stormpath tenant api key secret'
}

var stormpathClient = new stormpath.Client({
  apiKey: tenantApiKey
});

var app = express();
var appHref = 'https://api.stormpath.com/v1/applications/3WIeKpaEjPHfLmy6GIvbwv';
var application;
var clientCredentialsAuthenticator;
var jwtAuthenticator;
var groupScopeFactory; // see previous example above

stormpathClient.getApplication(appHref, function(err, application) {

  if (err) {
    throw err;
  }

  clientCredentialsAuthenticator = new stormpath.OAuthClientCredentialsAuthenticator(application);
  clientCredentialsAuthenticator.setScopeFactory(groupScopeFactory);
  clientCredentialsAuthenticator.setScopeFactorySigningKey(tenantApiKey.secret);

  jwtAuthenticator = new stormpath.JwtAuthenticator(application);

  // Now that we have authenticators, we can start the server
  app.listen();

});


app.post('/oauth/token', function (req, res, next){
  if (req.body.grant_type !== 'client_credentials') {
    return next(new Error('Unsupported grant type'));
  }

  // This client ID and secret is an API key that has been
  // issued to the end-user's stormpath account, this is not
  // a Tenant API Key


  var authenticationRequest = {
    apiKey: {
     id: req.body.client_id,
     secret: req.body.client_secret
    },
    scope: req.body.scope;
  };

  clientCredentialsAuthenticator.authenticate(authenticationRequest, function(err, authenticationResult) {
    if (err) {
      return next(err);
    }

    res.json(authenticationResult.accessTokenResponse);
  });
});

function requireAdmins(req,res,next) {
  var accessToken; // Pull the access token from the `Authorization: Bearer <token>` location

  jwtAuthenticator.authenticate(accessToken, function (err, authenticationResult) {
    if (err) {
      return next(err);
    }

    if (authenticationResult.accessToken.expandedJwt.claims.scope.match(/admin/)) {
      return next(); // Authorized, OK to proceed
    }

    res.status(403).end();
  });
}

app.get('/admins', requireAdmins, function (req, res, next) {
  res.send('You are an admin');
});
```

Tweety-FER:
Implements the scope factory feature for certain OAuth authenticators. This enables developers to attach functions to their authenticators, with the goal of attaching `scope` properties to their JSON Web Tokens (JWT), and as such their authentication results. These can then be used for various purposes, as described in the attached issue.

Adds tests and class documentation. Modifies some existing classes slightly to satisfy the requirements.

Usage example:

``` js
var application; // instantiated somewhere
var apiKey; // fetched beforehand
var stormpathToken; // generated beforehand
var authenticator;
var jwtAuthenticator;

function groupScopeFactory(authenticationResult, requestedScope, callback){
  authenticationResult.getAccount({expand: 'groups'}, function(err, account){
    if (err) {
      return callback(err);
    }

    var groupNames = account.groups.items.map(function (group) {
      return group.name;
    });

    var grantedScope = groupNames.join(' ');

    callback(null, grantedScope);
  });
}

jwtAuthenticator = new stormpath.JwtAuthenticator(application);

authenticator = new stormpath.OAuthStormpathTokenAuthenticator(application);
authenticator.setScopeFactory(groupScopeFactory);
authenticator.setScopeFactorySigningKey(apiKey.secret);

authenticator.authenticate({stormpath_token: stormpathToken, scope: 'admin'}, function(err, authResult) {
  authenticator.authenticate(authResult.accessTokenResponse.access_token, function(err, authConfirm) {
    // Will be defined if the user is a member of at least one group
    console.log(authConfirm.expandedJwt.claims.scope);
  });
});
```

Fixes #556

### DIFF
--- a/lib/jwt/jwt-authentication-result.js
+++ b/lib/jwt/jwt-authentication-result.js
@@ -104,7 +104,7 @@ function JwtAuthenticationResult(application,data) {
    * @type {Jwt}
    */
 
-  if(this.accessToken) {
+  if(this.accessToken){
     this.accessToken = nJwt.verify(this.accessToken, apiKey.secret);
     this.account = {
       href: this.accessToken.body.sub
@@ -123,7 +123,7 @@ function JwtAuthenticationResult(application,data) {
    * @type {Jwt}
    */
 
-  if(this.refreshToken) {
+  if(this.refreshToken){
     this.refreshToken = nJwt.verify(this.refreshToken, apiKey.secret);
   }
 }
--- a/lib/jwt/jwt-authentication-result.js
+++ b/lib/jwt/jwt-authentication-result.js
@@ -61,9 +61,9 @@ var nJwt = require('njwt');
  * @param {AccessTokenResponse} accessTokenResponse
  * The access token response from the Stormpath REST API.
  */
-function JwtAuthenticationResult(application,data) {
+function JwtAuthenticationResult(application, data) {
   if (!(this instanceof JwtAuthenticationResult)) {
-    return new JwtAuthenticationResult(application,data);
+    return new JwtAuthenticationResult(application, data);
   }
 
   /**
@@ -79,11 +79,11 @@ function JwtAuthenticationResult(application,data) {
     on this object - but convert underscores to camelcase because
     that's the node way bro.
    */
-  Object.keys(data).reduce(function(a,key){
-    var newKey = key.replace(/_([A-Za-z])/g, function (g) { return g[1].toUpperCase(); });
+  Object.keys(data).reduce(function(a, key) {
+    var newKey = key.replace(/_([A-Za-z])/g, function(g) { return g[1].toUpperCase(); });
     a[newKey] = data[key];
     return a;
-  },this);
+  }, this);
 
   /*
     Assign application after the key reduction above,
@@ -104,7 +104,7 @@ function JwtAuthenticationResult(application,data) {
    * @type {Jwt}
    */
 
-  if(this.accessToken){
+  if (this.accessToken) {
     this.accessToken = nJwt.verify(this.accessToken, apiKey.secret);
     this.account = {
       href: this.accessToken.body.sub
@@ -123,7 +123,7 @@ function JwtAuthenticationResult(application,data) {
    * @type {Jwt}
    */
 
-  if(this.refreshToken){
+  if (this.refreshToken) {
     this.refreshToken = nJwt.verify(this.refreshToken, apiKey.secret);
   }
 }
--- a/lib/jwt/jwt-authentication-result.js
+++ b/lib/jwt/jwt-authentication-result.js
@@ -61,9 +61,9 @@ var nJwt = require('njwt');
  * @param {AccessTokenResponse} accessTokenResponse
  * The access token response from the Stormpath REST API.
  */
-function JwtAuthenticationResult(application, data) {
+function JwtAuthenticationResult(application,data) {
   if (!(this instanceof JwtAuthenticationResult)) {
-    return new JwtAuthenticationResult(application, data);
+    return new JwtAuthenticationResult(application,data);
   }
 
   /**
@@ -79,11 +79,11 @@ function JwtAuthenticationResult(application, data) {
     on this object - but convert underscores to camelcase because
     that's the node way bro.
    */
-  Object.keys(data).reduce(function(a, key) {
-    var newKey = key.replace(/_([A-Za-z])/g, function(g) { return g[1].toUpperCase(); });
+  Object.keys(data).reduce(function(a,key){
+    var newKey = key.replace(/_([A-Za-z])/g, function (g) { return g[1].toUpperCase(); });
     a[newKey] = data[key];
     return a;
-  }, this);
+  },this);
 
   /*
     Assign application after the key reduction above,
@@ -104,7 +104,7 @@ function JwtAuthenticationResult(application, data) {
    * @type {Jwt}
    */
 
-  if (this.accessToken) {
+  if(this.accessToken) {
     this.accessToken = nJwt.verify(this.accessToken, apiKey.secret);
     this.account = {
       href: this.accessToken.body.sub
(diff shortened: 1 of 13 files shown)


robertjd:
Thanks @Tweety-FER , my apologies for taking so long to review this.  Two items to discuss:

(1) In this PR, it appears that the scope factory does not receive an `authenticationResult` as the first argument.  I'm getting back a JWT, which I trace back to here:

https://github.com/stormpath/stormpath-sdk-node/pull/558/files#diff-4e11d153b6051b0635dd983b48a2d536R126

(2) The `requestedScope` value is only needed by the scope factory.  It's to allow the developer to know what scopes the user requested, it's then up to the developer to decide what scopes to give.  This scope value shouldn't be posted to our REST API, only provided to the scope factory function.

Let me know, thanks!

Tweety-FER:
@robertjd Not really much to discuss. 1) was a mistake on my part, so fixed that. As for 2), wasn't aware of that, but it does make sense not to send it up to the API. Removed it now.

Again, the ITs seem to be failing for no particular reason I can detect, and in spots unrelated to this PR's code, in different environments for repeated runs. It seems to be a mix of environmental and possibly test setup issues.

Tweety-FER:
Note that I am looking into these IT failures that seem to be plaguing us recently. Will attempt to resolve them in a separate PR.

Tweety-FER:
@robertjd Did further tweak for 1) as we discussed today, and updated the tests. I think this should be good to go now.

robertjd:
Thanks @Tweety-FER , I do get an `authenticationResult` now.  But when I try to run the specified `groupScopeFactory` example, I run into a new problem: `JwtAuthenticationResult.getAccount()` does not support expansions, we should fix that in this PR.  Did you not run into this problem when trying to run this example?

Tweety-FER:
@robertjd Right you are, I actually slightly tweaked the example I ran, writing the tweaked in the description without actually testing it. Ensured that it run now, with slight changes to the description, as well. Sorry about that.

In worse news, some of the unrelated ITs _still_ seem to be failing, even after rebasing on `master` and including the PR that should have made some of it better. Will have to get back to that.

robertjd:
Done!  I made another pass at the docs, and setup a few more classes that needed to inherit from this new class, thanks @Tweety-FER!

PR #566, opened 2016-11-14
Integration test fixes

Tweety-FER:
We've recently had some integration test failures that happen only on occasion (due to timeout issues, tests not handling async properly, etc).

This PR is an attempt to resolve these errors in two ways:
  - In certain cases, these issues are resolved, or the probability of them happening lowered
  - In other cases, error reporting is added/improved

### DIFF
--- a/test/it/oauth_authenticator_it.js
+++ b/test/it/oauth_authenticator_it.js
@@ -75,14 +75,16 @@ describe('OAuthAuthenticator',function(){
               }else{
                 application2 = app2;
                 /*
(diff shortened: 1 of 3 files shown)


Tweety-FER:
Note that I'm making no guarantees that this solves *everything*. However, I did go through the last several build failures and tried to track the issues down. At the very least, the frequency of these errors should reduce, and I can keep on investigating any future build failures.

robertjd:
Thanks @Tweety-FER !  This looks great and is really helpful.  As you say, we can investigate future failures if they happen.

PR #570, opened 2016-11-16
Update collection methods

### LINKED ISSUE #569: Filtering collections isn't working properly
Code:

```
          account.getGroupMemberships({expand: 'group'}, (err, memberships) => {
            memberships.filter((membership, next) => {
              console.log('checking if membership should be included', membership.group.name);
              const test = membership.group.name.startsWith('myString');
              console.log('will be included?', test);
              next(null, test);
            }, (err, memberships) => {
              console.log('got filtered group memberships', err, memberships);
            });
          });
```

Output:
```
checking if membership should be included myString:831f9e62cd
will be included? true
checking if membership should be included myString:831f9e62cd/example:34a800114c
will be included? true
checking if membership should be included stuff:831f9e62cd/example:34a800114c
will be included? false
got filtered group memberships [] undefined
```

I would expect `err` to be `null` and `memberships` to contain the two first memberships. Documentation: https://docs.stormpath.com/nodejs/jsdoc/CollectionResource.html#filter__anchor

Using stormpath 0.18.5 on node 6.9.1.

Tweety-FER:
Update the `async` library to `2.1.x` and update the code and tests to reflect this.

The most important changes are that some methods that previously took `next` functions and `callback`s that only had an arity of one now have an arity of 2. For example, `filter` can now be called as:

```javascript
collection.filter(function(item, next) {
  if (item.id < 0) {
     return next(err);
  }

  next(null, item.id % 2 === 0);
}, function(err, filtered) {
  if (!err) {
    console.log('Filtered:', filtered);
  }
});
```

The following methods (method aliases given in brackets) are affected:

- `filter` (`select`)
- `reject`
- `detect`
- `some` (`any`)
- `every` (`all`)

Fixes #569 

Makes these PRs redundant:
- #499
- #505

### DIFF
--- a/lib/resource/Application.js
+++ b/lib/resource/Application.js
@@ -1225,10 +1225,16 @@ Application.prototype.setDefaultAccountStore = function setDefaultAccountStore(s
       return callback(err);
     }
 
-    res.detectSeries(function(asm, cb){cb(asm.accountStore.href === store.href);}, onAsmFound);
+    res.detectSeries(function(asm, cb){
+      cb(null, asm.accountStore.href === store.href);
+    }, onAsmFound);
   });
 
-  function onAsmFound(asm) {
+  function onAsmFound(err, asm) {
+    if (err) {
+      throw err;
+    }
+
     if (asm) {
       asm.isDefaultAccountStore = true;
       return asm.save(clearCache);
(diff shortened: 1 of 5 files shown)


Tweety-FER:
It's our old friend, the `Jwt has expired` error.

robertjd:
Thanks @Tweety-FER !  I've created a new `1.0.0` branch, and changed this PR to use that branch as the base.  Let's start making any breaking change PRs into that branch, with the eventual goal to release a next major version.

I've tried running the `express-stormpath` tests while linked to these SDK changes, and most of the tests are failing.  Can you take a look at the cause?  I'd like to ensure that we've captured all places in this SDK where we need to update our async calls, and the `express-stormpath` tests are a good way to find places we don't have covered in the SDK tests

Tweety-FER:
@robertjd Thanks for pointing that out. Updated the example in this repo and fiddled with the express lib to see what's going on. After pointing the dep at this branch and fixing a couple of lines of code in the express lib, the errors apparently went away. Is it the same for you? It's this PR: https://github.com/stormpath/express-stormpath/pull/546

PR #576, opened 2016-11-29
Simplify test to avoid Jwt Expired

Tweety-FER:
Simplified the existing test as to avoid the persistent `Jwt Expired` error we keep getting when running the tests (even though everything is fine). 

After some testing, this looks like it fixes the issue, and this now matches the local authorisation tests that verify the same thing.


robertjd:
Good catch, thanks!

PR #584, opened 2016-12-14
Support groups collection request cache

cz-Michael:


robertjd:
Thanks @cz-Michael , we'll take a look at this (probably after this week as we have some other work right now).  We need to ensure that this change won't introduce unexpected behavior for those developers who rely on collections not being cached, for reasons of consistency.

PR #586, opened 2016-12-19
Fix/crypto deprecation

### LINKED ISSUE #541: pbkdf2 deprecation warning when calling application.getApiKey()
When using Node v6.5.0 and calling application.getApiKey(...) the following deprecation warning is written to console:

```
(node:29446) DeprecationWarning: crypto.pbkdf2 without specifying a digest is deprecated. Please specify a digest
```
#### How to reproduce
1. Use Node version v6.5.0.
2. Run the application below:
   
   ``` javascript
   
   var applicationHref = 'https://api.stormpath.com/v1/applications/YOUR_APP_HREF';
   var apiKeyId = 'YOUR_ACCOUNT_API_KEY_ID';
   
   client.getApplication(applicationHref, function (err, application) {
       if (err) {
           return console.error(err);
       }
   
       application.getApiKey(apiKeyId, function (err, apiKey) {
           if (err) {
               return console.error(err);
           }
   
           console.log('Application retrieved! Deprecation warning should be written before this.');
       });
   });
   ```
3. Check the console for the following deprecation warning:
   
   ```
   (node:29446) DeprecationWarning: crypto.pbkdf2 without specifying a digest is deprecated. Please specify a digest
   ```

Tweety-FER:
Updates `crypto.pbkdf2` call to use an explicit digest, to prevent a Node v6+ deprecation warning. It uses `sha1`, which used to be the default value in previous versions, so behaviour will remain unchanged.

Added a special `try-catch` to support Node `0.10.*`, because its version of `crypto` does not support the `digest` parameter at all. It can freely be removed once we dump `0.10.*`.

Fixes #541

### DIFF
--- a/lib/resource/ApiKey.js
+++ b/lib/resource/ApiKey.js
@@ -47,7 +47,9 @@ ApiKey.prototype._getDecryptedSecret = function _getDecryptedSecret(callback) {
   var encryptedSecret = new Buffer(this.secret,'base64');
   var iv = encryptedSecret.slice(0,16);
   var rawEncryptedValue = encryptedSecret.slice(16);
-  crypto.pbkdf2(password,new Buffer(salt,'base64'),iterations,(keyLengthBits/8), 'sha1',function(err,key){
+  var saltBuf = new Buffer(salt,'base64');
+  var keyLength = keyLengthBits / 8;
+  var cryptoCallback = function(err, key){
     if(err){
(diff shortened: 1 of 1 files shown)


shikloshi:
+ 1

alexpx:
+1

robertjd:
Awesome, thanks @Tweety-FER !

PR #591, opened 2017-01-06
Collection resource update

mdeggies:
Fixed broken collection resource method examples and removed docs for the methods that don't exist (collection.concatLimit(), collection.detectLimit(), collection.everyLimit(), collection.everySeries(), collection.filterLimit(),collection.mapLimit(), collection.rejectLimit(), collection.someLimit(), collection.someSeries(),collection.sortByLimit(), collection.sortBySeries()). 

One thing to note, there are still some methods remaining that don't work as expected (all of the limit() methods, like collection.eachLimit() and collection.mapLimit()). The docs say they're supposed to be exactly like their parent methods but with an iterator, so for ex, eachLimit should look like this: 

```
    function iterator(account, next) {
      console.log('Found account for ' + account.givenName + ' (' + account.email + ')');
      next();
    }

    function doneCallback(err) {
      if (!err) {
        console.log('All accounts have been visited.');
      }
    }

    application.getAccounts({ email: 'foo@example.com' }, function (err, collection) {
      if (!err) {
        collection.eachLimit(iterator, 2, doneCallback);
      }
    });
```

When I run this, though, it doesn't make it into any of the functions (iterator or doneCallback). Please take a look and let me know if they should be fixed here.

### DIFF
--- a/lib/resource/CollectionResource.js
+++ b/lib/resource/CollectionResource.js
@@ -268,28 +268,9 @@ utils.inherits(CollectionResource, Resource);
  *
  * application.getAccounts({ expand: 'customData' }, function (err, collection) {
  *   if (!err) {
- *     collection.concat(iterator, donCallback);
+ *     collection.concat(iterator, doneCallback);
  *   }
  * });
- */
-
-/**
- * @name CollectionResource#concatLimit
- *
- * @function
- *
- * @description
- * The same interface as `concat`, but with a 2nd parameter that can be used to
- * limit the number of parallel workers.
- *
- * @param {Function} iterator
- * Same interface as {@link CollectionResource#concat concat}.
- *
- * @param {Number} limit
- * The maximum number of parallel iterator workers.
- *
- * @param {Function} doneCallback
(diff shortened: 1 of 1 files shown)


Tweety-FER:
@mdeggies Sorry to not have reacted sooner (I was on vacation at the time this issue was raised), but we already have a PR that solves this issue, the other way around. See https://github.com/stormpath/stormpath-sdk-node/pull/570

The issue is that the documentation uses async v2, while the code uses async v1. The references PR fixes that, updating the lib to async v2 as a dependency. However, it's a breaking change, so it's going into `1.0.0.`.

Since this is a broader change (additional documentation changes are contained within), I'm just putting this comment here for reference, as the documentation will have to be reverted back to async v2 in the referenced PR if this is merged.

mdeggies:
@Tweety-FER ahh no problem! I'll leave this to @robertjd or Robin to see what they want to do. Just wanted to clean this up for the time being as we were getting a lot of questions about these methods not working as documented. If we don't merge, I think it would be good to at least add a note to the top of the page explaining the issue.

PR #597, opened 2017-01-23
Access token authenticator

robertjd:
This adds a new access token authenticator, which decouples application assertions from access token assertions (authentication vs. authorization).  It also adds a key resolver, which depends on a not-yet-released version of nJwt (work in progress here https://github.com/jwtk/njwt/pull/43).

This authenticator will be used with the next patch release of express-stormpath, so that access tokens issued by client API will validate correctly - so long as they are issued by the same stormpath application, we will still retain that behavior, we just need to introduce the key resolving.

### DIFF
--- a/lib/cache/CacheHandler.js
+++ b/lib/cache/CacheHandler.js
@@ -11,7 +11,7 @@ var utils = require('../utils');
 
 var CACHE_REGIONS = ['applications', 'directories', 'accounts', 'groups',
   'groupMemberships', 'tenants', 'accountStoreMappings','apiKeys','idSiteNonces',
-  'customData', 'organizations', 'authTokens'];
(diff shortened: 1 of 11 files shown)


lib/cache/CacheHandler.js
```diff
@@ -79,6 +80,11 @@ function getCacheByHref(cacheManager, href) {
     region = href.match(/customData/) ? 'customData' : (href.split('/').slice(-2)[0]);
   }
 
+  // if (href && href.match(/application\/[^\/]+\/authTokens\//)) {
```
nbarbettini: Remove commented lines?

typerandom:
I made some formatting changes, but other than that, I'm 👍  for this!

PR #609, opened 2017-06-23
Okta

skhvan1111:
Allows user to save arrays or any other objects as account customData.
When i tried to save array ["a", "b","c"] as customData.values
dot.dot produced 
"values.0": "a",
"values.1": "b",
"values.2": "c"
As its not valid fields set for okta, it raises 400 error with related message.


robertjd:
Hi @skhvan1111 , thanks for the PR!  We need to support deep objects as well, so this has to recurse downward.

Today I was fixing some other issues with custom data, and I have addressed the array issue described here.  Can you please try this commit on the head of the okta branch and let me know if it fixes your issues:

https://github.com/stormpath/stormpath-sdk-node/commit/bb4b28d51537ef423614c7c97543acd4a5e16477

Thanks!

